PR #10, opened 2021-11-12
Added sample files to handle file upload and download in endpoints

kyostiebi:
I added new controller to handle basic file upload and download operations.


garrettjstevens (changes requested):
I think the functionality looks like it works well. I have a few stylistic and error handling points to discuss:

- I think "file-handling" in the file names and endpoint should be changed to "fileHandling," to be consistent with other JBrowse projects
- I don't think any of the try-catch blocks are necessary. Unless we are trying to catch a specific error, I think it's best to let the app-level error handling deal with unexpected errors.
- We should throw exceptions instead of returning errors, again to let the app-level error handling deal with things in a consistent way. So instead of
```js
return res.status(HttpStatus.INTERNAL_SERVER_ERROR).json({
  status: HttpStatus.INTERNAL_SERVER_ERROR,
  message: 'File ' + filename + ' does not exist!',
})
```
do
```js
throw new InternalServerErrorException(
  'File ' + filename + ' does not exist!',
)
```

PR #11, opened 2021-11-16
Add .env file support, get db config from it

garrettjstevens:
@kyostiebi as discussed, this adds support for reading settings from a `.env` file. It uses the built-in `@nestjs/config` handling of env files. In development mode it will read from `.development.env`, and when not in development mode it will read from `.env`.

There is the ability using this method to use different formats for the config file if we need something more structured, and also to do validation of the config file as well if we want to.


kyostiebi (approved):
Looks good

kyostiebi (approved):

PR #35, opened 2022-01-25
Add a change manager

### LINKED ISSUE #27: Implement ChangeManager
- ChangeManager accepts ChangeObjects (#16)
- ChangeManager runs frontendPreValidation methods for all registered validations (for first pass assume no registered validations)
- ChangeManager applies the change to the client datastore
- ChangeManager runs frontendPostValidation methods for all registered validations (for first pass assume no registered validations)
  - If any validation fails, the change is reverted in the client datastore and a notification is shown to the user
- ChangeManager sends the change to the Driver

garrettjstevens:
This adds a change manager to the Apollo view model and has a demo UI that allows you to submit a change to the change manager that change a feature location's "end" (two buttons that show up under the view and allow you to increase/decrease the end of one specific feature).

Also includes the first implementation of a concrete Change: https://github.com/GMOD/jbrowse-plugin-apollo/blob/ebf15700c824ce3662df984207cb82a2be5dad2d/packages/jbrowse-plugin-apollo/src/ChangeManager/Change.ts

Resolves #27


cmdcolin:
can you add more context about this PR? it is not very clear what types of things this is similar to in e.g. apollo 1. is this a method for smoothing over the inmemory vs database systems? it seems like in some ways, this extra abstraction is a lot of work and the linfo is a bit hard to decode as an outsider (i'm not tuned into the apollo conversations but i'm curious)

garrettjstevens:
> can you add more context about this PR?

In Apollo, each update to the data is going to be made by creating an object for that change based on a common Change class. Each class will contain the logic for how to apply that change in different contexts, e.g. on the front-end data store or on the back-end database. The changes are also serializable so they can be shared between front and back ends and also written to a change log.

The change manager in this PR is in the client side and manages applying a change to the client data store and then sending it off to whatever back end is being used (e.g. local file or collaboration server).

PR #39, opened 2022-01-31
Initial commit for LoginWidget

kyostiebi:
Please, find LoginWidget branch for reviewing. In model.ts there is one
// eslint-disable-next-line @typescript-eslint/no-explicit-any 
that I was not able to get rid of.

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
+++ b/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
@@ -28,22 +28,17 @@ export const ApolloLoginForm = ({
       const queryParams = new URLSearchParams(paramsString)
       url.search = queryParams.toString()
 
-      fetch(url.toString(), {
+      const res = await fetch(url.toString(), {
         method: 'POST',
       })
-        .then((res) => res.json())
-        .then((data) => {
-          // If authentication was successfull then there is key 'token'
-          if ('token' in data) {
-            const responseToken = data.token
-            handleClose(responseToken)
-          } else {
-            alert('Authentication failed')
-          }
-        })
-        .catch((rejected) => {
-          alert(rejected)
-        })
+      const data = await res.json()
+      // If authentication was successfull then there is key 'token'
+      if ('token' in data) {
+        const responseToken = data.token
+        handleClose(responseToken)
+      } else {
+        alert('Authentication failed')
+      }
     } else {
       handleClose()
     }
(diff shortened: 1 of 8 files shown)


packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+  internetAccountId: string
+  handleClose: (arg?: string) => void
+}) => {
+  const [username, setUsername] = useState('')
+  const [password, setPassword] = useState('')
+
+  async function onSubmit(
+    this: unknown,
+    event: React.FormEvent<HTMLFormElement>,
+  ) {
```
garrettjstevens, line 26: The unused `this` parameter isn't necessary, and it's probably best to call `event.preventDefault()` at the top of the function to make sure it runs.

```suggestion
  async function onSubmit(event: React.FormEvent<HTMLFormElement>) {
    event.preventDefault()
```

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+          if ('token' in data) {
+            const responseToken = data.token
+            handleClose(responseToken)
+          } else {
+            alert('Authentication failed')
+          }
+        })
+        .catch((rejected) => {
+          alert(rejected)
+        })
```
garrettjstevens, line 48: Generally we try to use `async/await` style promises instead of `.then()` style promises for consistency.

```suggestion
      const res = await fetch(url.toString(), {
        method: 'POST',
      })
      const data = await res.json()
      // If authentication was successfull then there is key 'token'
      if ('token' in data) {
        const responseToken = data.token
        handleClose(responseToken)
      } else {
        alert('Authentication failed')
      }
```

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  async function onSubmit(
+    this: unknown,
+    event: React.FormEvent<HTMLFormElement>,
+  ) {
+    if (username && password) {
+      const url = new URL(authenticationURL)
+      const paramsString = `username=${username}&password=${password}`
+      const queryParams = new URLSearchParams(paramsString)
+      url.search = queryParams.toString()
```
garrettjstevens, line 31: It's probably better to have the username/password in the POST body instead of the URL, so it would look something like:

```js
const res = await fetch(authenticationURL, {
  method: 'POST',
  body: JSON.stringify({ username, password }),
  headers: { 'Content-Type': 'application/json' },
})
```

(In my testing, it needed the 'Content-Type' header to work properly.)

garrettjstevens (changes requested):

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/components/ApolloLoginForm.tsx
kyostiebi: Done as proposed

garrettjstevens (approved):

PR #42, opened 2022-02-08
Issue40

kyostiebi:
Ready for comments. Before you test functionality please take backup copy of volvox.sort.fasta.gff3 because new feature will add apollo_id to each feature (and overwrites original volvox.sort.fasta.gff3)


packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+  async checkGFF3uniqueKey(filenameWithPath: string) {
+    const stringOfGFF3 = await fs.readFile(filenameWithPath, {
+      encoding: 'utf8',
+      flag: 'r',
+    })
+    this.logger.verbose(`Data read from file=${stringOfGFF3}`)
+
+    const arrayOfThings = gff.parseStringSync(stringOfGFF3, {
+      parseAll: true,
+    })
```
garrettjstevens, line 525: I was wrong when we were talking earlier, this is right. We want `parseAll` so that the sequence, etc. ends up back in the file.

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+      // Comment, Directive and FASTA -entries are not presented as an array
+      if (Array.isArray(entry)) {
+        for (const [key, val] of Object.entries(entry)) {
+          this.logger.verbose(
+            `GFF3Item: key=${JSON.stringify(key)}, value=${JSON.stringify(
+              val,
+            )}`,
+          )
+          if (val.hasOwnProperty('attributes')) {
+            const assignedVal = Object.assign(val)
```
garrettjstevens, line 539: A type should be specified here to TypeScript knows what the type is.

```suggestion
            const assignedVal: GFF3FeatureLineWithRefs = Object.assign(val)
```

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+            // Let's add apollo_id to parent feature if it doesn't exist
+            if (!assignedVal.attributes.hasOwnProperty('apollo_id')) {
+              assignedVal.attributes.apollo_id = uuidv4()
+            }
```
garrettjstevens, line 543: This was the way of doing it we came up with earlier that TypeScript could understand better.
```suggestion
            if (!('apollo_id' in attributes)) {
              attributes.apollo_id = [uuidv4()]
            }
            assignedVal.attributes = attributes
```

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+        let i = 0;
+        i < Object.keys(parentFeature.child_features).length;
+        i++
+      ) {
+        this.logger.verbose(
+          `Child no #${i} has value=${JSON.stringify(
+            parentFeature.child_features[i][0],
+          )}`,
+        )
+        const assignedVal = Object.assign(parentFeature.child_features[i][0])
```
garrettjstevens, line 596: This should have a type, same as above. The lines below will need to be updated, too.

```suggestion
        const assignedVal: GFF3FeatureLineWithRefs = Object.assign(
          parentFeature.child_features[i][0],
        )
```

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+        // Let's add apollo_id if it doesn't exist yet
+        if (!assignedVal.attributes.hasOwnProperty('apollo_id')) {
+          assignedVal.attributes.apollo_id = uuidv4()
+        }
+        for (const k in assignedVal) {
+          if (
+            typeof parentFeature[k] == 'object' &&
+            parentFeature[k] !== null &&
+            parentFeature[k].length !== undefined &&
+            parentFeature[k].length > 0
```
garrettjstevens, line 606: What is this checking for?

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
```diff
@@ -503,4 +508,107 @@ export class FileHandlingService {
...
+      gff.formatSync(arrayOfThings),
+    )
+  }
+
+  /**
+   * Loop child features in parent feature and add apollo_id to each child's attribute
+   * @param parentFeature - Parent feature
+   */
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  setApolloIdRecursively(parentFeature: any) {
```
garrettjstevens, line 576: Should be `GFF3FeatureLineWithRefs` instead of `any`

```suggestion
  setApolloIdRecursively(parentFeature: GFF3FeatureLineWithRefs) {
```

garrettjstevens (changes requested):
I noticed that if there are multiple lines with the same ID (which is allowed by the GFF3 spec), only one of them gets assigned an ID. You should be able to replicate it with this minimal GFF3:

```
ctgA	example	gene	1050	9000	.	+	.	ID=EDEN;Name=EDEN;Note=protein kinase
ctgA	example	mRNA	1050	9000	.	+	.	ID=EDEN.1;Parent=EDEN;Name=EDEN.1;Note=Eden splice form 1;Index=1
ctgA	example	CDS	1201	1500	.	+	0	Parent=EDEN.1;ID=abcd
ctgA	example	CDS	1301	1600	.	+	0	Parent=EDEN.1;ID=abcd
```

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
kyostiebi: This is now removed. It was leftover from my previous tries when I occasionally got Undefined exception.

packages/apollo-collaboration-server/src/fileHandling/fileHandling.service.ts
kyostiebi: ok

kyostiebi:
Ready for new review

garrettjstevens (approved):

PR #71, opened 2022-04-05
Create Feature schema and add GFF3 loading

### LINKED ISSUE #59: Create Feature schema and GFF3 loading
- Update the schema in `packages/apollo-schemas/src/feature.schema.ts` to the current schema.
- Create a Feature module that has one endpoint, `features/importGFF3`
  - This will take the GFF3 text as a file upload in the body of a POST request
  - The assembly will also be specified in the POST request
  - e.g. the request might look like this: `curl http://localhost:3999/features/importGFF3 -F 'file=@./volvox.gff3' -F 'assembly=assemblyId'
- Make sure refSeq ID is validated with `mongoose-id-validator`, and that the right combination of assembly and refSeq are chosen

kyostiebi:
Closes #59


packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -0,0 +1,215 @@
...
+            this.logger.verbose(
+              `FeatureId assigned (recursive level) ${JSON.stringify(
+                assignedVal,
+              )}`,
+            )
+            featureIdArrAsParam.push(uid)
+          }
+          this.setAndGetFeatureIdRecursively(assignedVal, featureIdArrAsParam)
+        }
+      }
```
garrettjstevens, line 211: This can be a lot more concise with array mapping. Also, no need to check for `featureId` since we don't support loading those from GFF3.

```suggestion
    // If there are child features
    if (parentFeature.child_features) {
      parentFeature.child_features = parentFeature.child_features.map(
        (childFeature) =>
          childFeature.map((childFeatureLine) => {
            const featureId = uuidv4()
            featureIdArrAsParam.push(featureId)
            const newChildFeature = { ...childFeatureLine, featureId }
            this.setAndGetFeatureIdRecursively(
              newChildFeature,
              featureIdArrAsParam,
            )
            return newChildFeature
          }),
      ) as GFF3FeatureLineWithRefsAndFeatureId[][]
```

packages/apollo-schemas/src/feature.schema.ts
```diff
@@ -1,22 +1,26 @@
...
-  gff3FeatureLineWithRefs: { type: JSON, required: true },
-})
-export interface FeatureDocument extends GFF3FeatureLineWithRefs, Document {
-  id: string
-  refSeqId: string
+  @Prop({ type: [String], required: true, index: true })
   featureId: string[]
+
+  @Prop({ type: JSON, required: true })
   gff3FeatureLineWithRefs: GFF3FeatureLineWithRefs
```
garrettjstevens, line 23: I think we should spread the feature into the  top of the schema, for example like this:
```ts
@Schema()
export class Feature implements GFF3FeatureLineWithRefs {
  @Prop({
    required: true,
    index: true,
    type: MongooseSchema.Types.ObjectId,
    ref: 'RefSeq',
  })
  refSeqId: RefSeq

  @Prop({ type: [String], required: true, index: true })
  featureId: string[]

  @Prop({ required: true })
  // eslint-disable-next-line camelcase
  seq_id: string

  @Prop()
  source: string

  @Prop({ required: true })
  type: string

  @Prop({ required: true })
  start: number

  @Prop({ required: true })
  end: number

  @Prop()
  score: number

  @Prop()
  strand: string

  @Prop()
  phase: string

  @Prop({ type: Map, of: [String] })
  attributes: Record<string, string[]>

  @Prop({ type: JSON })
  // eslint-disable-next-line camelcase
  child_features: GFF3Feature[]

  @Prop({ type: JSON })
  // eslint-disable-next-line camelcase
  derived_features: GFF3Feature[]
}
```

packages/apollo-schemas/src/feature.schema.ts
```diff
@@ -1,22 +1,26 @@
...
-export const FeatureSchema = new Schema({
-  refSeqId: { type: String, required: true },
-  featureId: { type: [String], required: true },
-  gff3FeatureLineWithRefs: { type: JSON, required: true },
-})
-export interface FeatureDocument extends GFF3FeatureLineWithRefs, Document {
-  id: string
-  refSeqId: string
+  @Prop({ type: [String], required: true, index: true })
   featureId: string[]
```
garrettjstevens, line 20: We should probably have two props, `featureId` which is the ID of the parent feature, and `feautureIds`, which is the IDs of all the child features.

garrettjstevens (changes requested):
I pushed a couple small changes to make the structure the same as our other modules. Other notes inline.

It also would be good for parent features that don't have any children to omit `child_features` and `derived_features` altogether instead of having them be an empty array.

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -0,0 +1,215 @@
...
+    let isFeature = false // Indicates if we need to get valid AssemblyId and refSeqId
+    // Loop all lines
+    for (const entry of arrayOfThings) {
+      // eslint-disable-next-line prefer-const
+      let featureIdArray: string[] = [] // Let's gather here all feature ids from each feature
+      isFeature = false
+
+      // Comment, Directive and FASTA -entries are not presented as an array
+      if (Array.isArray(entry)) {
+        isFeature = true
```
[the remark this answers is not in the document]
kyostiebi: The purpose of 'isFeature' was to separate features from comments, directives and sequences which will be also saved into database but using different refSeqId. So, later in code when we are saving comments/directives/sequences into database we need to find refseqId separately.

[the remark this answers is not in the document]
kyostiebi: I removed isFeature -boolean flag. I'll will insert similar logic later when it will be needed.

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -0,0 +1,215 @@
...
+          currentSeqId = assignedVal.seq_id!
+          this.logger.verbose(
+            `Added new FeatureId: key=${JSON.stringify(
+              key,
+            )}, value=${JSON.stringify(val)}`,
+          )
+          // Check if there is also childFeatures in parent feature and it's not empty
+          if (
+            val.hasOwnProperty('child_features') &&
+            Object.keys(assignedVal.child_features).length > 0
```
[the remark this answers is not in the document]
kyostiebi: Double check removed

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -0,0 +1,215 @@
...
+        for (const [key, val] of Object.entries(entry)) {
+          // Let's add featureId to parent feature if it doesn't exist
+          const assignedVal: GFF3FeatureLineWithRefsAndFeatureId =
+            Object.assign(val)
+          const uid = uuidv4()
+          assignedVal.featureId = uid
+          // Add featureId into array
+          featureIdArray.push(uid)
+          // Pick up refSeq (i.e. seq_id)
+          currentSeqId = assignedVal.seq_id!
```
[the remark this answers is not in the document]
kyostiebi: Fixed as proposed

packages/apollo-collaboration-server/src/features/features.controller.ts
```diff
@@ -0,0 +1,36 @@
...
+   * or in case of error return throw exception
+   */
+  @Post('/importGFF3')
+  @UseInterceptors(FileInterceptor('file'))
+  async importGFF3(
+    @UploadedFile() file: Express.Multer.File,
+    @Body() body: AssemblyIdDto,
+  ) {
+    this.logger.debug(`Adding new features for assemblyId: ${body.assemblyId}`)
+    const fileName = await this.featuresService.saveNewFile(file)
```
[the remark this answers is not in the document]
kyostiebi: Fixed.

packages/apollo-schemas/src/feature.schema.ts
kyostiebi: If we spread feature up to top level we have to think what shall we populate into those mandatory fields when we are inserting comments and directives into Features -collection in database.

packages/apollo-schemas/src/feature.schema.ts
kyostiebi: ParentId will be also in featureIds array

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -0,0 +1,215 @@
...
+      isFeature = false
+
+      // Comment, Directive and FASTA -entries are not presented as an array
+      if (Array.isArray(entry)) {
+        isFeature = true
+        this.logger.verbose(`ENTRY=${JSON.stringify(entry)}`)
+        for (const [key, val] of Object.entries(entry)) {
+          // Let's add featureId to parent feature if it doesn't exist
+          const assignedVal: GFF3FeatureLineWithRefsAndFeatureId =
+            Object.assign(val)
```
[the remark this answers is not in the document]
kyostiebi: Fixed

packages/apollo-collaboration-server/src/features/features.service.ts
kyostiebi: Fixed as proposed.

packages/apollo-schemas/src/feature.schema.ts
kyostiebi: Moved to top level.

packages/apollo-schemas/src/feature.schema.ts
kyostiebi: Done. Parent feature id is now saved in _parentFeatureId_ -property

garrettjstevens (approved):

PR #73, opened 2022-04-05
Add range fetching endpoint

kyostiebi:
This branch covers features that are listed in https://github.com/GMOD/jbrowse-plugin-apollo/issues/62

### DIFF
--- a/packages/apollo-collaboration-server/src/features/features.service.ts
+++ b/packages/apollo-collaboration-server/src/features/features.service.ts
@@ -106,13 +106,18 @@ export class FeaturesService {
       parseAll: true,
     })
     let cnt = 0
+    let currentSeqId = ''
+    let refSeqId = ''
+    let isFeature = false // Indicates if we need to get valid AssemblyId and refSeqId
     // Loop all lines
     for (const entry of arrayOfThings) {
       // eslint-disable-next-line prefer-const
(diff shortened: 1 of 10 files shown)


packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -362,4 +363,49 @@ export class FeaturesService {
...
+        'gff3FeatureLineWithRefs.start': { $lte: searchDto.end },
+        'gff3FeatureLineWithRefs.end': { $gte: searchDto.start },
+        seq_id: refSeqIdIdArray,
+      })
+      .exec()
+    this.logger.debug(
+      `Searching features for AssemblyId: ${searchDto.assemblyId}, refName: ${searchDto.refName}, start: ${searchDto.start}, end: ${searchDto.end}`,
+    )
+
+    if (!features) {
```
garrettjstevens, line 400: Will this ever be undefined? If there are no matches, it should be an empty array, right?

kyostiebi: As background:
**_The findOne() returns first document if query matches otherwise returns null. The find() method does not return null, it returns a cursor_**
So, after we make query we have to check if return value was NULL or in case of find() -method, we have to check the length of cursor (i.e. is it less than one = no docs found)

garrettjstevens:
Closing in favor of #74

Issue #67, opened 2022-03-31
Make a way to load test data into MongoDB


garrettjstevens:
Create some way (possible a `.mongodb` playground file) that allows you to load all necessary data for one organism into an empty MongoDB instance. That includes Assembly, RefSeqs, RefSeqChunks, and Features. This will allow us to load test data without relying on the UI.

kyostiebi:
@garrettjstevens now, there is also schema definition and test data for refSeqChunks -collection.

PR #76, opened 2022-04-11
Add refSeqChunks schema and module

### LINKED ISSUE #66: Create RefSeqChunks schema and module
Create an RefSeqChunks schema in `apollo-schemas` and module (using nest g resource assemblies), but then clear out the controller of most methods since they'll be tied to the RefSeqs module endpoints. Make sure the service methods are filled out, though.

garrettjstevens:
Closes #66

Does not include loading sequence from FASTA or GFF3 yet, but we'll be able to load sequences when a PR implementing #67 is merged.


garrettjstevens:
@kyostiebi I need to look over this one more time before merging, but the schemas in this branch should be good to go for you to use when making test data.

PR #78, opened 2022-04-12
Add first applyToServer method for Changes

### LINKED ISSUE #63: Add applyToServer to LocationEndChange and LocationStartChange
- Add abstract `applyToServer` to `packages/apollo-shared/src/ChangeManager/Change.ts`
  - Also includes adding new ServerDataStore interface
- Implement `applyToServer` in LocationEndChange and LocationStartChange so that they work with the MongoDB backend

kyostiebi:
Ready for review

Closes #63


kyostiebi:
New draft PR contains update logic only for End -position. Once End -position logic has passed review, I'll copy it to Start -position.

packages/apollo-collaboration-server/src/features/features.controller.ts
```diff
@@ -75,4 +77,15 @@ export class FeaturesController {
...
+  /**
+   * Updates end position of given feature. Before update, current end -position value is checked (against given old-value)
+   * @param serializedChange - Information containing featureId, newEndValue, oldEndValue
+   * @returns Return 'HttpStatus.OK' if featureId was found AND oldEndValue matched AND database update was successfull. Otherwise throw exception.
+   */
+  //  @UseGuards(JwtAuthGuard)
+  @Put('/updateEndPos')
+  updateMongo(@Body() serializedChange: SerializedChange) {
+    return this.featuresService.changeEndPos(serializedChange)
+  }
```
garrettjstevens, line 90: This will need to be moved to `packages/apollo-collaboration-server/src/change/change.controller.ts`. Probably what needs to happen is that the login in `submitChange` in `change.controller.ts` needs to be moved to `change.service.ts`, and then we need to inject `featureModel` into the `ChangeService`.

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -35,10 +41,31 @@ export class FeaturesService {
...
+      await change.apply({
+        typeName: 'Server',
+        featureModel: this.featureModel,
+      })
+    } catch (error) {
+      throw error
+    }
+    return []
+  }
+
```
garrettjstevens, line 68: As noted above, this should be part of `submitChange` in `ChangeService` and not in `FeatureService`.

packages/apollo-shared/src/ChangeManager/LocationEndChange.ts
```diff
@@ -42,6 +45,80 @@ export class LocationEndChange extends Change {
     }
   }
 
+  // ****** JOS KÄYTTÄÄ NESJS/COMMONia niin tulee luultavasti RUNTIME ERROR **********
+  /**
+   * Applies the required change to database
+   * @param backend - parameters from backend
+   * @returns
+   */
+  async applyToServer(backend: ServerDataStore) {
```
garrettjstevens, line 54: The code for finding the (sub)feature by ID can be simplified, see `findById` in `features.service.ts` in the current main branch, or this commit: b54c932811096b49d4c078705cea926e7f1a9dd8

packages/apollo-shared/src/ChangeManager/LocationEndChange.ts
```diff
@@ -42,6 +45,80 @@ export class LocationEndChange extends Change {
...
+    const { changes } = this
+    // eslint-disable-next-line prefer-destructuring
+    const { featureId, oldEnd: expectedOldEnd, newEnd } = changes[0]
```
garrettjstevens, line 57: This should be done for every change, not just the first one.

garrettjstevens (changes requested):
I've re-based this branch so that it's simpler to compare to main. I also updated some things so that they match the code in main better, and restored local GFF3 editing so that this PR is only adding new functionality and not yet removing any of the existing functionality. Some comment inline.

packages/apollo-shared/src/ChangeManager/LocationEndChange.ts
garrettjstevens: In fact, we should probably create a `FeatureChange` class that extends `Change`, and put the code for finding a feature by ID in that class. Then `LocationEndChange` and `LocationStartChange` can both extend `FeatureChange` and share that code.

packages/apollo-collaboration-server/src/features/features.controller.ts
kyostiebi: End-point has now been moved to change.controller.ts.

packages/apollo-shared/src/ChangeManager/LocationEndChange.ts
garrettjstevens: Created `FeatureChange` and updated the `findById`.

garrettjstevens:
@kyostiebi this is ready for you to add the transaction-type checks you mentioned earlier. One thing to not in my changes is that I added back the local GFF3 editing code so that it still works and the MongoDB code is adding new code instead of replacing existing code. Having both in place will help with the UI transition.

kyostiebi:
We need to add proper Mongo transaction handling to endpoint to ensure data consistency in case of multiple changes. It means that we commit changes to database only if all changes are successful. If any change fails then we need to rollback all changes.

garrettjstevens:
@kyostiebi I pushed some changes that you can pull and check out. For handling the connection like we talked about earlier today, I found it was easiest to create the session in `change.service.ts` and then pass the session in to the change in `apollo-shared`. I also created the session with the `transaction` helper, which takes care of calling `commitTransaction` (or `abortTransaction` if an error is thrown) for us, so we don't have to explicitly call them.

I also made it so it saves the features that it finds when it first queries the database so it doesn't have to do a second query after it checks all the start/ends.

garrettjstevens:
@kyostiebi I fixed the MongoDB transaction error by removing `session.endSession()`. It turns out the `transaction` helper wrapper already runs that for us.

garrettjstevens (approved):

PR #96, opened 2022-05-16
Add assembly from file

kyostiebi:
Ready for draft PR. Please, check line 140 in AddAssemblyFromFileChange.ts. I was not able to use CHUNK_LEN environment variable in match() -method.


garrettjstevens:
I had to rebase this branch to make it work with the last PR's changes, so you'll want to do `git reset --hard origin/Issue92` before continuing work on this branch.

packages/apollo-shared/src/ChangeManager/AddAssemblyFromFileChange.ts
```diff
@@ -0,0 +1,176 @@
...
+              // Add refSeqChunks
+              const newRefSeqChunkDoc = await refSeqChunkModel.create({
+                refSeq: newRefSeqDoc._id,
+                n: ind,
+                sequence: chunk,
+              })
+              this.logger.debug?.(
+                `Added new refSeqChunk (n=${ind}) docId "${newRefSeqChunkDoc._id}" for refSeq "${newRefSeqDoc._id}"`,
+              )
+              ind++
```
garrettjstevens, line 152: > I was not able to use CHUNK_LEN environment variable in match() -method.

I'd probably do something like this instead of using `match()`.

```suggestion
            const chunkLength = Number(CHUNK_LEN)
            const numChunks = Math.ceil(data.sequence.length / chunkLength)
            for (let chunkNum = 0; chunkNum < numChunks - 1; chunkNum++) {
              const start = chunkNum * chunkLength
              const chunk = data.sequence.slice(start, start + chunkLength)
              await refSeqChunkModel.create({
                refSeq: newRefSeqDoc._id,
                n: chunkNum,
                sequence: chunk,
              })
            }
```

garrettjstevens:
I added a test file `packages/apollo-collaboration-server/test/data/volvox.fa`. This is a FASTA file, which should also work with this endpoint (as well as a GFF3 file).

PR #98, opened 2022-05-23
Issue97

kyostiebi:
There is still issue with Mongoose transaction in AddAssemblyFromFileChange.ts in lines 106 and 204


garrettjstevens:
I made sure all db actions used the session, and I think the errors with transactions are gone now.

There is one issue I noticed where if the stream data ends partway through a line and the line is a definition line, it cuts off the definition and adds the description to the sequence. You can replcate this by adding a small `highWaterMark` to the read stream, like this:
```js
const sequenceStream = fs
  .createReadStream(compressedFullFileName, { highWaterMark: 12 })
  .pipe(createGunzip())
```

cmdcolin:
what are streams being used for? my ears perk up when I see stream stuff just because I know it is tricky to do right

garrettjstevens:
Streams are being used to process FASTA files for adding sequence information to the database.

cmdcolin:
could you refer to a FAI file for this? we have faidx-js for creating a FAI file from a FASTA if needed https://github.com/GMOD/faidx-js

garrettjstevens:
We need the sequence itself as well, so we can't use FAI only.

cmdcolin:
Why do you need the sequence? Could also adopt the streaming parser in there to your needs if interested. But, are you storing the sequence in the database? Why not refer to a external fasta file from the DB?

garrettjstevens:
We are storing the sequence in the database, since one of our requirements is that we need to eventually be able to edit the sequence, which we decided will be easier to do in the database.

cmdcolin:
Just as a data point. Apollo 1 allows sequence edits (stored as deltas) and stores files outside the DB. Not saying Apollo 1 is entirely correct, but it's a point of reference.

Further more: doing queries for slices of the reference sequence is likely faster with indexed fasta than a large text blob in the database. With indexed fasta, only a small part of the sequence is read. With a large text string stored in the database, it might read this entire chromosome into memory ever time you want to do a small sequence read

Possible pros and cons here https://dba.stackexchange.com/a/226464/117914

garrettjstevens:
We're not storing each chromosome as a single blob. We're storing them as chunks, using a method inspired by MongoDB's [GridFS](https://www.mongodb.com/docs/manual/core/gridfs/).

cmdcolin:
to me this sounds more complicated than using a fasta file on disk, but i will defer to you guys!

PR #100, opened 2022-05-24
Create a change for adding both assembly and features

kyostiebi:


garrettjstevens:
I think factoring out the `addFeatureToDb` change is a good idea. It would probably be better to add it as a method in `FeatureChange.ts` instead, though, and other changes can extend that abstract class. Also, then you can pass in just `feature` and `backend` to the method, since the other params will be available on the class.

kyostiebi:
`addFeatureToDb` method has now been moved to `FeatureChange.ts` class. Also the parameters have been changed accordingly. Changes pushed.

garrettjstevens (changes requested):
I rebased this so it will apply cleanly to the main branch, you'll want to `git reset --hard origin/Issue99` before continuing work.

Right now `AddAssemblyAndFeaturesFromFileChange` adds an assembly, but not the `refSeq`s from that assembly. It should add those `refSeq`s, the same as `AddAssemblyFromFileChange` does. Some of that assembly/refseq code can probably be put in the `FeatureChange` class.

kyostiebi:
Now adds also RefSeqs into Mongo.

garrettjstevens:
@kyostiebi It looks like there aren't any changes on this branch. Maybe your `git push` didn't work?

garrettjstevens (changes requested):
`AddAssemblyFromFileChange` should also use `addRefSeqIntoDb` the same as `AddAssemblyAndFeaturesFromFileChange` does.

PR #102, opened 2022-05-27
Issue101

kyostiebi:
Ready for first review


packages/jbrowse-plugin-apollo/src/components/AddAssembly.tsx
```diff
@@ -0,0 +1,115 @@
...
+  async function onSubmit(event: React.FormEvent<HTMLFormElement>) {
+    event.preventDefault()
+    const sessionToken = sessionStorage.getItem('apolloInternetAccount-token')
+    if (sessionToken == null) {
+      alert('You must authenticate first!')
+      return
+    }
+    console.log(`Assembly name is "${assemblyName}"`)
+
+    const res = await fetch(new URL('/changes/submitChange', baseURL).href, {
```
garrettjstevens, line 44: Instead of getting the token manually, you'll need to get the custom fetcher from the internet account. That would look something like this:

```js
const uri = new URL('/changes/submitChange', baseURL).href
const apolloFetch = apolloInternetAccount.getFetcher({
  locationType: 'UriLocation',
  uri,
})
const res = await apolloFetch(uri, { //...
```

That way you don't have to add the authorization  header, it'll get added for you automatically.

packages/jbrowse-plugin-apollo/src/components/stateModel.ts
```diff
@@ -0,0 +1,20 @@
+import { types, Instance } from 'mobx-state-tree'
```
garrettjstevens, line 1: I don't think you'll need this stateModel file

garrettjstevens (changes requested):
You'll need to add a field in the dialog for the user to select their file. Then when the user submits, there should be two requests:
1) A request to `/files` to add the file, getting back the file object
2) A request to `changes/submitChange` to create an assembly from the file that just got added.

kyostiebi:
Added Radio button list to choose file type. And added BSON to generate new assemblyId in UI.

garrettjstevens (changes requested):
Let's have the "submit" button be disabled until all required fields are filled. Also, there should be some error handling so that if one of the `fetch`es gives an error, the error text is displayed in the dialog for the user.

kyostiebi:
Submit button is now disabled until fields are filled. Error messages shown to user in case of error in fetches

garrettjstevens (approved):

PR #123, opened 2022-07-15
Add "View change log" dialog

kyostiebi:
Ready for test. There are a couple issues:
1. Drop-down (select) menus. I was not able to set each item per row even I have tried different flexDirection -options.
2. I was not able to set DataGrid column adjustable (in terms of width)

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,9 +45,5 @@
   "resolutions": {
     "typescript": "^4.7.4"
   },
-  "private": true,
-  "dependencies": {
-    "@mui/icons-material": "^5.8.4",
-    "react-collapsed": "^3.3.2"
-  }
+  "private": true
 }
--- a/package.json
+++ b/package.json
@@ -45,5 +45,8 @@
   "resolutions": {
     "typescript": "^4.7.4"
   },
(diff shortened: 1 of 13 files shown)


garrettjstevens:
For (1), there is some sort of conflict between the Accordion and the Select. It might be easier to remove the Accordion.
For (2), it looks like the adjustable width feature is only available in the pro version of DataGrid

I tried making some changes that would make (1) not a problem anymore. Basically, I added a toolbar to the data grid that allows the users to choose the filters from there. By making the columns have the right types (e.g. "Change type" is a `singleSelect`, "createdAt" is a `dateTime`), the filtering is taken care of for us. The build-in toolbar also adds a couple other nice options, like hiding columns and exporting the data.

You can see my branch here: [117_changelog_gs](https://github.com/GMOD/Apollo3/tree/117_changelog_gs). You'll need to use the [more_grid_exports](https://github.com/GMOD/jbrowse-components/tree/more_grid_exports) branch of jbrowse-components to get that branch to work. Here's what it looks like:

image

garrettjstevens (approved):

PR #144, opened 2022-10-06
Add role-based authorization

kyostiebi:
Ready for draft PR. Before doing test, please add for example the following documents into users -collection in Mongo.
```json
[
{
  "_id": {
    "$oid": "632d5d1025e16ef49ce75667"
  },
  "id": 1,
  "email": "demo@ebi.ac.uk",
  "role": "readOnly",
  "username": "demo"
},
{
  "_id": {
    "$oid": "632d5d9625e16ef49ce75669"
  },
  "username": "john",
  "id": 2,
  "email": "pomo@ebi.ac.uk",
  "role": [
    "user"
  ]
}
]
```


garrettjstevens (commented):
Could you generate a full "users" resource that we can use? You can do this with
```
cd packages/apollo-collaboration-server/src/
yarn nest g resource users
```
and then moving the relevant code from `usersDemo` to `users` (I think we can get rid of the demo users in the users service, too).

Also, could you remove the `any`s?

garrettjstevens (changes requested):
I've pushed a couple small changes, so you'll want to pull those down before continuing.

- AuthorizationValidation will need to be moved to apollo-shared, e.g. `packages/apollo-shared/src/Validations/AuthorizationValidation.ts`
- `AuthorizationValidation` should be a class that extends the `Validation` base class, and the logic that is right now in `backendPreValidateAuthorize` should be in the `backendPreValidate` method of that class
- If you need to, you can add a second parameter to `backendPreValidate` in both `Validation` and `ValidationSet` to pass anything else you might need, e.g. `usersService`
- You'll need to import and register the `AuthorizationValidation` in `main.ts`
- In the `ValidationGuard`, instead of calling `backendPreValidateAuthorize`, you should import `validationRegistry` from `apollo-shared` and call `validationRegistry.backendPreValidate`

garrettjstevens:
As discussed earlier, you can ignore the first bullet point above for now.

garrettjstevens (approved):

PR #145, opened 2022-10-07
Predict uploaded assembly file type based on selected file extension

haessar:


garrettjstevens (changes requested):
Looks good, just a couple things:

- I've noticed if I choose a .fasta file, the bottom checkbox gets disabled, but stays checked. It should get disabled and get un-checked.
- The second thing is a bit more of an abstract idea, since the change I'd suggest doesn't change the behavior of the UI. It's more of a way of thinking about React. What I'd suggest is doing the below, which is basically turning the radio buttons into a [controlled component](https://reactjs.org/docs/forms.html#controlled-components). That way the logic is centralized in the RadioGroup instead of each individual FormControl.

```diff
diff --git a/packages/jbrowse-plugin-apollo/src/components/AddAssembly.tsx b/packages/jbrowse-plugin-apollo/src/components/AddAssembly.tsx
--- a/packages/jbrowse-plugin-apollo/src/components/AddAssembly.tsx
+++ b/packages/jbrowse-plugin-apollo/src/components/AddAssembly.tsx
@@ -207,16 +207,17 @@ export function AddAssembly({
               defaultValue="text/x-gff3"
               name="radio-buttons-group"
               onChange={handleChangeFileType}
+              value={fileType}
             >
               <FormControlLabel
                 value="text/x-gff3"
-                control={<Radio checked={fileType === 'text/x-gff3'} />}
+                control={<Radio />}
                 label="GFF3"
                 disabled={submitted && !errorMessage}
               />
               <FormControlLabel
                 value="text/x-fasta"
-                control={<Radio checked={fileType === 'text/x-fasta'} />}
+                control={<Radio />}
                 label="FASTA"
                 disabled={submitted && !errorMessage}
               />
```

PR #146, opened 2022-10-11
Issue 142: User management panel with dummy data

haessar:


packages/jbrowse-plugin-apollo/src/components/ManageUsers.tsx
```diff
@@ -0,0 +1,115 @@
...
+  handleClose(): void
+}
+
+export function ManageUsers({ session, handleClose }: ManageUsersProps) {
+  const { internetAccounts } = getRoot(session) as AppRootModel
+  const [errorMessage, setErrorMessage] = useState('')
+
+  const users = useUsers(internetAccounts, setErrorMessage)
+  type Row = typeof users[number]
+  const [rows, setRows] = useState<Row[]>(users)
```
garrettjstevens, line 41: React's behavior here might not be intuitive, this will set `rows` to whatever `users` is the first time the code runs, but won't update it, so it will always be an empty array. You could probably use `users` directly without an intermediate `rows` state.

garrettjstevens:
In response to comments on https://github.com/GMOD/Apollo3/issues/149

> I've bulked out all the changeManager infrastructure for the ManageUsers panel. The async function onChangeTableCell is correctly triggering when changing the role dropdown value, and the params from onCellEditCommit contains the necessary information to apply the change, but for some reason I continue to see the message Uncaught (in promise) TypeError: changeManager is undefined in the browser console. I'm not getting any linting issues, and I've added it to ManageUsersProps etc in the same way as seen in AddAssembly.tsx, so I'm a bit stumped

You'll need to add it to the props when you do `queueDialog`, like here: https://github.com/GMOD/Apollo3/blob/98d7833832fd9d3a185e3c7c4f0c56694ac8d562/packages/jbrowse-plugin-apollo/src/index.ts#L139-L149

> On another note, the requirement for an assemblyId in all Change class descendants feels a bit redundant with User model changes (as assemblies don't come into it). Is it worth making a parallel Change class that duplicates the functionality of the original Change class without that parameter, or should that parameter requirement be moved to a child of the Change class?

I've got a branch I'm working on where Changes are split up more so that not all of them need assemblyIds. For now you can just pass any string for assemblyId.

haessar:
Will try to quickly hook up the delete button with a destroy request now.

haessar:
Also, this 
image
provides motivation to include the userId as a Change attribute at the same level as assemblyId currently is (or however it will be with your in-progress code)

haessar:
I've brought the getUsers function into the ManageUsers component as a callback, to allow me to perform useEffect to initialise the user data.

packages/jbrowse-plugin-apollo/src/components/ManageUsers.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export function ManageUsers({
+  session,
+  handleClose,
+  changeManager,
+}: ManageUsersProps) {
+  const { internetAccounts } = getRoot(session) as AppRootModel
+  const apolloInternetAccounts = internetAccounts.filter(
+    (ia) => ia.type === 'ApolloInternetAccount',
+  ) as ApolloInternetAccountModel[]
```
garrettjstevens, line 52: Should also probably check here that ia.role has 'admin'

garrettjstevens (changes requested):
I've merged the branch for #162 in here, so that now the new changes in the branch don't need to provide an assembly or changedIds.

One thing I've noticed is that when deleting a user, it doesn't update in the UI unless you close and open the dialog again.

packages/jbrowse-plugin-apollo/src/components/ManageUsers.tsx
haessar: fixed the role change ui bug and added the check for admin role. Noticed what looks like a bug in the roles hierarchy being returned in the jwt token from the authentication service:

```
// Loop user's role(s) and add each role + inherited ones to userRolesArray
    for (const userRole of userFound.role) {
      const roles = RoleInheritance[userRole] // Read from role.enum.ts
      roles.forEach((role) => {
        userRoles.add(role)
      })
    }

    const payload = {
      username: userFound.username,
      email: userFound.email,
      roles: Array.from(userRoles),
    }
```
This all makes sense on paper, but I've tested by changing my "admin" account to "readOnly", verified the role has updated in the db, and still the roles array in the jwt token shows all 3 roles (when only [Role.ReadOnly] should be there according to the RoleInheritance). Could this be a caching issue?

garrettjstevens: I think I see what's going on. If you change the role for yourself, until you log out and log in again, you'll still have your old role in the jwt.

Could you update this so that you can't make changes to your own account? Still show the row in the manage users screen, but have the edit/delete functionality disabled. This way we can guarantee that an instance doesn't end up without an admin, and I think it's pretty standard practice in account management.

haessar: OK I've made the suggested change and it appears to be working as desired now. I added the user id to the jwt payload so that I can retrieve it in the same way I'm doing with the role in ManageUsers component.

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
```diff
@@ -63,6 +63,14 @@ const stateModelFactory = (
         const dec = jwtDecode(token) as JWTPayload
         return dec.roles
       },
+      get userId() {
```
haessar, line 66: This is probably not the most efficient way to do this as I'd assume retrieveToken is relatively expensive operation which would become apparent when many users are in the db.

haessar:
I've rebased onto main and made necessary tweaks to ensure it's still working

garrettjstevens (approved):

PR #147, opened 2022-10-11
Add role info to JWT payload

kyostiebi:
Ready for draft PR

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloRenderer/components/ApolloRendering.tsx
+++ b/packages/jbrowse-plugin-apollo/src/ApolloRenderer/components/ApolloRendering.tsx
@@ -379,7 +379,7 @@ function ApolloRendering(props: ApolloRenderingProps) {
           Copy features and annotations
         </MenuItem>
         <MenuItem
-          disabled={!isAdmin}
+          // disabled={!isAdmin}
           key={3}
           value={3}
           onClick={() => {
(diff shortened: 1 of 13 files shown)


kyostiebi:
Merged changes from Issue136 and got rid of lint warnings

packages/jbrowse-plugin-apollo/src/ApolloRenderer/components/ApolloRendering.tsx
```diff
@@ -63,6 +65,22 @@ function ApolloRendering(props: ApolloRenderingProps) {
...
   )
+
+  useEffect(() => {
+    const { internetAccounts } = getRoot(session) as AppRootModel
+    const apolloInternetAccount = internetAccounts.find(
+      (ia) => ia.type === 'ApolloInternetAccount',
+    ) as ApolloInternetAccountModel | undefined
+    if (!apolloInternetAccount) {
+      throw new Error('No Apollo internet account found')
+    }
```
garrettjstevens, line 76: We have to make sure the internetAccount is the correct one, instead of the first one. You'll need to use the assemblyName from the region and the assemblyManager to get the internetAccountConfigId for the assembly and use that to find the correct account. An example of that is here: https://github.com/GMOD/Apollo3/blob/0ee1d3f9781b468215483fa59ff108827c44c112/packages/apollo-shared/src/BackendDrivers/CollaborationServerDriver.ts#L21-L40

garrettjstevens (changes requested):
I rebased on main and added a couple changes. The biggest thing was making `roles` a view instead of state on the internetAccount model. When it was in state, it was getting erased on page refresh.

One more change that needs to be made, noted inline.

garrettjstevens (approved):

PR #156, opened 2022-11-02
Add increment to changes collection

haessar:
All the infrastructure seems to be there and passing linting

### DIFF
--- a/packages/apollo-collaboration-server/src/counters/counters.service.ts
+++ b/packages/apollo-collaboration-server/src/counters/counters.service.ts
@@ -1,0 +1,30 @@
+import { Injectable, Logger, NotFoundException } from '@nestjs/common'
+import { InjectModel } from '@nestjs/mongoose'
+import { Counter, CounterDocument } from 'apollo-schemas'
+import { Model } from 'mongoose'
+
+@Injectable()
+export class CountersService {
+  constructor(
+    @InjectModel(Counter.name)
+    private readonly counterModel: Model<CounterDocument>,
+  ) {}
+
+  private readonly logger = new Logger(CountersService.name)
+
+  async getNextSequenceValue(sequenceName: string): Promise<number> {
+    const sequenceDocument = await this.counterModel
+      .findOneAndUpdate({
+        filter: { _id: sequenceName },
+        update: { $inc: { sequenceValue: 1 } },
+        returnNewDocument: true,
+      })
+      .exec()
+    if (!sequenceDocument) {
+      const errMsg = `ERROR when getting next sequence value`
+      this.logger.error(errMsg)
+      throw new NotFoundException(errMsg)
+    }
+    return sequenceDocument.sequenceValue
+  }
+}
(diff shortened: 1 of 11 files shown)


haessar:
Having an issue when applying changes on top of latest main branch:

ERROR [ValidationGuard] TypeError: Cannot destructure property 'userModel' of 'undefined' as it is undefined.

I'm assuming this is related to the latest auth additions. If you've encountered it, anything I can do to easily fix?

garrettjstevens:
You might try rebasing on main again. I just did that and didn't see any error.

I did notice, though, that if there is not an entry in the sequences collection yet (e.g. after Apollo is first set up and there aren't any changes), then it gives an error. So you'll need to add some code to initialize the sequences collection if it's empty.

haessar:
@garrettjstevens I've rebased, checked there are sequences in db, restarted, rebuilt the remote container etc but still getting issues.
image
Currently seeing **TypeError: AuthPlugin.exports is undefined** in the browser when attempting to load jbrowse_config (see screenshot). I can see that the **AuthPlugin = pluginManager.getPlugin('AuthenticationPlugin')** was added to model.ts yesterday as part of PR 159.

haessar:
This is proving a bit of a blocker, so help on this asap would be appreciated.

haessar:
I've also tried running "yarn" in the root directories of the apollo-collaboration-server and jbrowse-plugin-apollo as I noticed the package.json deps were updated there

garrettjstevens:
@haessar sorry for the delay, I was away for a few days. It looks like you need to pull the latest changes to main in `jbrowse-components`, some of the auth stuff depends on some recent changes there.

packages/apollo-collaboration-server/src/counters/counters.service.ts
```diff
@@ -0,0 +1,30 @@
...
+@Injectable()
+export class CountersService {
+  constructor(
+    @InjectModel(Counter.name)
+    private readonly counterModel: Model<CounterDocument>,
+  ) {}
+
+  private readonly logger = new Logger(CountersService.name)
+
+  async getNextSequenceValue(sequenceName: string): Promise<number> {
```
haessar, line 15: This function isn't behaving. It's successfully finding the sequence value but seemingly not incrementing the value when changes are made. I've attempted all alternatives for "returnNewDocument" outlined here https://stackoverflow.com/a/47918879 (e.g. "returnOriginal", etc), all to no avail.

haessar: Found the issue, I was calling findOneAndUpdate incorrectly (passing all parameters as a dictionary in the first slot).

packages/apollo-collaboration-server/src/counters/counters.service.ts
```diff
@@ -0,0 +1,30 @@
...
+    const sequenceDocument = await this.counterModel
+      .findOneAndUpdate(
+        { id: sequenceName },
+        { $inc: { sequenceValue: 1 } },
+        { new: true, upsert: true },
```
garrettjstevens, line 20: I had to add "upsert" here, which inserts a new document if there isn't one yet, to make it work on a fresh database.

PR #170, opened 2022-11-23
Add communication over websockets

kyostiebi:
Ready for review


packages/apollo-shared/src/ChangeManager/ChangeManager.ts
```diff
@@ -86,9 +94,96 @@ export class ChangeManager {
     }
   }
 
+  async submitToClientOnly(change: Change) {
```
garrettjstevens, line 97: Instead of making a new `submitToClientOnly`, this should use the existing `submit(change, {submitToBackend: false})`

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
```diff
@@ -389,3 +404,127 @@ export type ApolloInternetAccountStateModel = ReturnType<
...
+    const token = internetAccount.retrieveToken()
+    if (!token) {
+      throw new Error(`No Token found`)
+    }
+    const { changeManager } = (session as ApolloSessionModel).apolloDataStore
+    socket.on('COMMON', (message) => {
+      // Save the last server timestamp
+      sessionStorage.setItem('LastSocketTimestamp', message.timestamp)
+      if (message.channel === 'COMMON' && message.userToken !== token) {
+        changeManager?.submitToClientOnly(message.changeInfo)
```
garrettjstevens, line 453: This should use the existing `submit`. You'll have to turn the changeInfo into a change first, which you can do with:

```ts
import { Change } from 'apollo-shared'

// ... 

const change = Change.fromJSON(message.changeInfo)
changeManager?.submit(message.changeInfo, { submitToBackend: false })

```

packages/apollo-collaboration-server/src/changes/changes.service.ts
```diff
@@ -105,11 +114,164 @@ export class ChangesService {
           `Error in backend post-validation: ${errorMessage}`,
         )
       }
+      this.logger.debug(`TypeName: ${change.typeName}`)
+
+      // Assembly related changes are broadcasted to 'COMMON' -channel
+      if (
+        change.typeName !== 'AddAssemblyAndFeaturesFromFileChange' &&
+        change.typeName !== 'AddAssemblyFromFileChange' &&
+        change.typeName !== 'DeleteAssemblyChange'
```
garrettjstevens, line 123: Instead of checking type names explicitly, this should check based on the Change class hierarchy. For example, if a `Change` is an `AssemblySpecificChange` but not a `FeatureChange`, then it should be broadcast on the common channel.

garrettjstevens (changes requested):

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
kyostiebi: Done

packages/apollo-collaboration-server/src/changes/changes.service.ts
kyostiebi: Done

packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
```diff
@@ -94,15 +94,16 @@ function AutocompleteInputCell(props: GridRenderEditCellParams) {
...
     const editable = useMemo(() => {
-      const { internetAccounts } = getRoot(session) as AppRootModel
-      const apolloInternetAccount = internetAccounts.find(
-        (ia) => ia.type === 'ApolloInternetAccount',
-      ) as ApolloInternetAccountModel | undefined
-      if (!apolloInternetAccount) {
-        throw new Error('No Apollo internet account found')
-      }
       return Boolean(apolloInternetAccount.getRole()?.includes('user'))
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
garrettjstevens, line 106: What was the reason for this change? I'd like to not change it if possible, since I don't want to run `find` on all accounts every render.

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
```diff
@@ -86,9 +99,81 @@ export class ChangeManager {
     }
   }
 
+  async changeThroughSocket(change: Change) {
```
garrettjstevens, line 102: What is this function for? It looks like we should be able to `change.apply` and not have to worry about re-creating each change when one already exists.

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
```diff
@@ -86,9 +99,81 @@ export class ChangeManager {
...
+    }
+
+    // submit to client data store
+    await ch?.apply(this.dataStore)
+  }
+
   async revert(change: Change, submitToBackend = true) {
-    const inverseChange = change.getInverse()
-    return this.submit(inverseChange, { submitToBackend, addToRecents: false })
+    change.getInverse()
```
garrettjstevens, line 176: What is this change to `revert` for?

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
```diff
@@ -389,3 +422,127 @@ export type ApolloInternetAccountStateModel = ReturnType<
 >
 export type ApolloInternetAccountModel =
   Instance<ApolloInternetAccountStateModel>
+
+/**
+ * Get server timestamp and save it into session storage
+ * @param apolloInternetAccount - apollo internet account
+ * @returns
+ */
+async function getAndSetServerTime(session: ApolloSession) {
```
garrettjstevens, line 431: This should get the sequence number of the most recent change instead of using a timestamp.

packages/jbrowse-plugin-apollo/src/ApolloRenderer/components/ApolloRendering.tsx
```diff
@@ -282,6 +296,16 @@ function ApolloRendering(props: ApolloRenderingProps) {
         bp: apolloFeatureUnderMouse[overEdge],
       })
     }
+    if (apolloFeatureUnderMouse && internetAccount) {
+      const userLocation = {
+        assemblyId: apolloFeatureUnderMouse.assemblyId,
+        refSeq: apolloFeatureUnderMouse.refSeq,
+        featureId: apolloFeatureUnderMouse._id,
```
garrettjstevens, line 303: We don't need a feature ID, only a location.

packages/jbrowse-plugin-apollo/src/session.ts
```diff
@@ -126,6 +237,23 @@ const ClientDataStore = types
...
+      // for (const view of (self as unknown as AbstractSessionModel).views) {
+      //   if (view.type === 'LinearGenomeView') {
+      //     console.log(`VIEW: ${JSON.stringify(view)}`)
+      //     // const {dynamicBlocks} = view
+      //     // // view and get location
+      //     // console.log(`BLOCKS: ${JSON.stringify(dynamicBlocks)}`)
+      //     // // run in https://developer.mozilla.org/en-US/docs/Web/API/setInterval
+      //   }
+      // }
+    },
```
garrettjstevens, line 256: This needs to be on the session, not the data store (e.g. above `apolloSetSelectedFeature`). This should work to get you the locations:

```ts
getLocations() {
  const locations = []
  for (const view of self.views) {
    if (view.type === 'LinearGenomeView') {
      const { dynamicBlocks } = view as LinearGenomeViewModel
      dynamicBlocks.forEach((block) => {
        if (block.regionNumber !== undefined) {
          const { assemblyName, refName, start, end } = block
          locations.push({ assemblyName, refName, start, end })
        }
      })
    }
  }
  return locations
}
```

garrettjstevens (changes requested):

packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
kyostiebi: I simply forgot to restore it yesterday after I did some tests. Sorry for that. There is no need to change anything in ApolloDetails.tsx

packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
kyostiebi: Fixed (wrong branch was pushed)

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
kyostiebi: I don't recognise this change either. Restored back.

packages/jbrowse-plugin-apollo/src/ApolloRenderer/components/ApolloRendering.tsx
kyostiebi: Unnecessary featureId removed.

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
kyostiebi: Fixed

packages/apollo-shared/src/ChangeManager/ChangeManager.ts
kyostiebi: Fixed

packages/jbrowse-plugin-apollo/src/session.ts
kyostiebi: Fixed

garrettjstevens (changes requested):
@kyostiebi, I realized we can't have the websocket stuff in the session, since it won't apply when we're editing a local GFF3. Also, we'll need to be able to handle sockets for multiple servers, so I've made `socket` an attribute on the apolloInternetAccountl, so each server will have it's own socket.

Here's what I need you to do:

- Move all general socket listeners (COMMON, USER_LOCATION, etc.) to `packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts` it looks like some of it is already duplicated there.
- Move region-specific socket listeners to `packages/apollo-shared/src/BackendDrivers/CollaborationServerDriver.ts`, in the `getFeatures` method. It's pretty much the same thing you did in `loadFeatures`, you'll check before the features are returned to make sure the refName has a listener. You'll also have access to the internetAccount there, so you can use `internetAccount.socket`.

garrettjstevens (approved):

Issue #174, opened 2022-12-09
Create Artemis-style six-frame feature view


garrettjstevens:
Create a new display type that has six row, one for each frame, and shows features with each subfeature on it's associated frame row. See the Artemis screenshot below for an example. Display will also likely include sequence integration, such as the marks in the below screenshot that indicate stop codons.

![image](https://user-images.githubusercontent.com/25592344/206591656-d8aa4a40-4842-4c3a-a08e-8a146a1f7e30.png)

haessar:
Note of a bug I encountered while testing out the existing different display types for the Reference sequence track: Attempting to choose the "GC content display" display results in a bunch of mobx-state-tree errors in the console. This happens both on the live demo and my dev server.

cmdcolin:
just a random idea: this type of feature could be made a part of jbrowse itself, as it could benefit non-apollo users. 



also note also how apollo 1 does this (it just uses colored boxes on different CDS subfeatures instead of this six frame feature) on both the annotation track and other gene feature tracks

![Screenshot from 2022-12-13 11-35-09](https://user-images.githubusercontent.com/6511937/207416877-bf87fa55-53ff-4fbb-849f-dd9aba413103.png)


see also https://github.com/GMOD/jbrowse-components/issues/3010

would also be curious what issue you saw @haessar

haessar:
This is the error I see both locally and in the demo when opening the "GC content display" display type in the Reference sequence track in Linear Genome View
![image](https://user-images.githubusercontent.com/11962461/207576963-658fa10b-fa0d-4846-9d6c-eeca5c47fd24.png)

cmdcolin:
@haessar nice work on the six frame prototype :) certainly it's always a design balance, probably just making multiple optiosn available is good

i made a possible fi for the GC content display issue here https://github.com/GMOD/jbrowse-components/pull/3402 if you want to try it out

PR #175, opened 2022-12-14
Add six frame feature display

haessar:
Very basic implementation.

I have used the LinearApolloDisplay as a basis for a new display: SixFrameFeatureDisplay (name expected to change). Most of the logic change is in the featureLayout method in stateModel.ts. Where before, new rows were added whenever features would potentially overlap, I am now tying the row number directly to the phase of each given feature. Naturally, as most features don't have a phase in the gff file, we end up with virtually all features plotted being CDS (but this is expected behaviour in Artemis), and I have to dig down two "generations" to find these CDS features for a given "gene" feature.

Still to do:
1) connecting lines between each CDS feature belonging to a common parent.
2) The central tracks representing forward/reverse strands (or at least a dividing horizontal line to represent the break between strands). Parent features might be plotted on these central tracks.
3) sequence integration (for stop codons, etc)

### DIFF
--- a/packages/apollo-common/src/Change.ts
+++ b/packages/apollo-common/src/Change.ts
@@ -19,6 +19,7 @@ export interface ClientDataStore {
   assemblies: Map<string, ApolloAssemblyI>
   internetAccounts: AppRootModel['internetAccounts']
   loadFeatures(regions: Region[]): void
+  loadRefSeq(regions: Region[]): void
   getFeature(featureId: string): AnnotationFeatureI | undefined
   addAssembly(_id: string, assemblyName: string): void
   addFeature(assemblyId: string, feature: AnnotationFeatureSnapshot): void
--- a/packages/apollo-mst/src/AnnotationFeature.ts
+++ b/packages/apollo-mst/src/AnnotationFeature.ts
@@ -57,6 +57,9 @@ export const AnnotationFeature = types
     get length() {
       return self.end - self.start
     },
+    get testId() {
+      return self.attributes.get('id')
+    },
     /**
      * Possibly different from `start` because "The GFF3 format does not enforce
      * a rule in which features must be wholly contained within the location of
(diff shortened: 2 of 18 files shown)


haessar:
I'm aware that I might need to rip this out again if we decide to go down the route of implementing it in base jbrowse2  (as suggested by @cmdcolin in #174), but hopefully can reuse some of the eventual logic.

garrettjstevens:
The prototype looks great! Looks like a solid base for working on your still-to-dos. I think there might be some opportunity to factor our some common code in the models of the two displays, but it's probably best to wait until things are more finalized to worry about that.

As for doing this in JBrowse, it would be good to keep in mind what parts could work in JBrowse, but we should keep developing this in Apollo for now. Right now the displays as we have them and their various mouse events, etc. are pretty tightly coupled with the Apollo-specific data and actions that the Apollo plugin manages in the session. Some of the layout and rendering code could potentially be separated out so it's not Apollo-dependent, though. We can discuss that in our next meeting.

haessar:
image
Forward codons are rendering correctly at all zoom levels. Still need to tweak the x coord or effective frame to get reverse codons to render correctly. Excuse all the messy commented out code. Will ensure it's tidied next week

haessar:
image

- All CDS features correctly rendering on reading frame without stop-codon overlap (I tried an elegant solution (i.e. feature start base divisibility by 3 mapping directly to row number) to this but had to resort to calling get codonLayout and checking for intersection with each prospective feature - at least this way ensures avoiding stop-codon overlap is "baked in").
- Stop/start codons are rendered correctly (using the jbrowse Reference sequence track as validation) on both forward and reverse strands. Stop/start codons (as well as feature intron lines) can be toggled on/off in context menu for track ("three-dots" menu).
- Never got round to having intron lines highlighted with mouse-over/selection (as is the case in Artemis) - instead kept the original implementation of distinct colours for common parent.
- Still notice on certain files (particularly larger) the getSequence call is erroneously throwing an exception when tracking along the genome - still not entirely clear to me why this is but I'm sure this will be something trivial in apollo-mst/src/index.txt:66

PR #185, opened 2023-02-03
Issue177 store and query ontologies

kyostiebi:
Current implementation includes the following:
- short instruction (in _ontologies.services.ts_) how OBO file can be converted to OBO-JSON file.
- a sample OBO JSON file has been added into (_./test/data/obo-converted.json_)
- UI makes typescript query against the file when user is adding a new child feature using _LinearApolloDisplay_

For now, the code also contains commented lines how data was added and queried using MongoDB (just in case we need to come back to it later stage). These commented lines can be removed if those are unnecessary.


garrettjstevens:
Closing, superseded by #188

PR #211, opened 2023-05-10
new glyphs mouse events

### LINKED ISSUE #208: Add mouse handling to customizable glyphs
This will start from the dynamic_blocks_rendering branch

There is currently an abstract Glyph class (`packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/glyphs/Glyph.ts`) that has `draw`, `getRowCount` and `getFeatureFromLayout` methods.

The glyph needs some way of reacting to mouse events, though. Currently the mouse handlers are defined in `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel.ts`, and there's state e.g. `apolloFeatureUnderMouse`, and then an autorun that watches that state and reacts to it (by e.g. drawing a mouseover highlight). Instead of drawing a mouseover highlight from the stateModel, though, we might need to keep not only the feature under the mouse, but the x and y positions, too, and then pass these into some method on the glyph and have it decide whether to draw something.

Specifically, the `'LinearApolloDisplayRenderMouseoverAndDrag'` autorun should be updated to call the glyph instead of drawing directly.

The actual canvases being rendered to are here: `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/LinearApolloDisplay.tsx`

Currently the only implemented glyph is the BoxGlyph, where every feature is only one line and subfeatures are not rendered.

rbuels:
fixes #208


rbuels:
@garrettjstevens this is getting close to mergeable into dynamic_blocks, the main thing remaining is to implement all the mouse stuff for the gene glyph

rbuels:
@garrettjstevens I’d recommend merging this into dynamic_blocks and going from there. There are quite a few new methods on the Glyph class for mouse events, and they are implemented for the Box glyph. The gene glyph isn’t implemented yet though because it wasn’t working on the data I was using

PR #212, opened 2023-05-18
Fix feature copying, improve conversion of GFF3 to AnnotationFeature

### LINKED ISSUE #193: Remove CopyFeatureChange and replace with AddFeatureChange
I was trying to fix `executeOnClient` in the `CopyFeatureChange`, and eventually realized that there are several things that don't work with this change. The biggest two problems are that (1) there isn't enough information in the change to be able to audit the change in the changelog and (2) all the reassigning of feature ids happens on the server, with no way to get that to the client to make the correct copy in the client data store.

What we'll need to do is remove `CopyFeatureChange`, and then the one place it's used we'll replace it with `AddFeatureChange`. So when a user copies a change, the client will copy the change, give the feature and subfeatures new IDs, change the refSeq id, and then use that info in an `AddFeatureChange`.

### LINKED ISSUE #197: More CopyFeatures features
After #193 is done, some additional features will be needed in the Copy Feature dialog:

- Be able to choose assembly that feature will be copied to
- Be able to choose refName that feature will be copied to, with the default being the same named ref, if one exists
- Be able to choose the location that the feature will be copied to
  - Validate that the location is valid before allowing submit

garrettjstevens:
Fixes #193
Fixes #197


cmdcolin:
any way to avoid storing the data files (go.json?) in the repository?

garrettjstevens:
> any way to avoid storing the data files (go.json?) in the repository?

It's not there anymore after rebase

PR #222, opened 2023-06-01
Add attributes column to details view

kyostiebi:
Ready for review

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
+++ b/packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
@@ -1,6 +1,6 @@
 import { AppRootModel, getSession } from '@jbrowse/core/util'
 import CloseIcon from '@mui/icons-material/Close'
-import { Autocomplete, IconButton, TextField } from '@mui/material'
+import { Autocomplete, Button, IconButton, TextField } from '@mui/material'
 import {
   DataGrid,
   GridColDef,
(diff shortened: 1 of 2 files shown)


packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
```diff
@@ -148,20 +185,78 @@ export const ApolloDetails = observer(
...
+
+    if (assemblyManager.get(assembly)?.refNameAliases) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+      const refNames = assemblyManager.get(assembly)!.refNameAliases!
+      Object.keys(refNames).forEach((key) => {
+        if (key === refSeq) {
+          refName = refNames[key]
+        }
+      })
+    }
```
garrettjstevens, line 199: There's a method on the assembly manager called `getCanonicalRefName` that we can use here.

```suggestion
    const { assemblyManager } = session

    const refName =
      assemblyManager.get(assembly)?.getCanonicalRefName(assembly) || refSeq
```

garrettjstevens (changes requested):
Looks good. A few adjustments:

- I noticed that if we use `columnVisibilityModel`, the built-in column show/hide doesn't work anymore. See ("Hide column" and "Manage columns" in screenshot).
image
I think if we remove the id column in `getFeatureColumns`, the other things that use the ID will still since it's still there internally.
- I'd like to avoid having an extra "Link to attributes" column. I think that we should open the attributes editor if someone double-clicks on the attributes cell, the same as if someone can edit the Start or End cells by double-clicking on them. I think we can do this by using the `onCellEditStart` event, and in the callback for that event we [manually stop the cell editing](https://mui.com/x/react-data-grid/editing/#controlled-mode) and open the attributes view instead.

packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/ApolloDetails.tsx
cmdcolin: small correction  `assemblyManager.get(assembly)?.getCanonicalRefName(refSeq) || refSeq`

garrettjstevens (approved):

PR #223, opened 2023-06-02
Fix for occasional server when adding or removing a feature

kyostiebi:
I tested this fix several times and I was not anymore able to re-produce any error even I fired automatically X times AddFeatureChange -change from AddFeature.tsx

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/components/AddFeature.tsx
+++ b/packages/jbrowse-plugin-apollo/src/components/AddFeature.tsx
@@ -85,7 +85,7 @@ export function AddFeature({
       }
       setPossibleChildTypes(data)
     }
-    getTypes()
+    getTypes().catch((e) => setErrorMessage(String(e)))
   }, [baseURL, internetAccount, sourceFeature.type])
 
   async function onSubmit(event: React.FormEvent<HTMLFormElement>) {
--- a/packages/jbrowse-plugin-apollo/src/components/AddFeature.tsx
+++ b/packages/jbrowse-plugin-apollo/src/components/AddFeature.tsx
@@ -110,7 +110,7 @@ export function AddFeature({
       },
       parentFeatureId: sourceFeature._id,
     })
-    changeManager.submit?.(change)
+    await changeManager.submit?.(change)
     notify(`Feature added successfully`, 'success')
     handleClose()
     event.preventDefault()
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -33,6 +33,7 @@ module.exports = {
       { selector: 'function', format: ['camelCase', 'PascalCase'] },
       { selector: 'typeLike', format: ['PascalCase'] },
     ],
+    '@typescript-eslint/no-floating-promises': 'error',
     '@typescript-eslint/no-shadow': 'error',
     curly: 'warn',
     camelcase: ['error', { properties: 'never', ignoreDestructuring: false }],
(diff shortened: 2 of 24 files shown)


packages/jbrowse-plugin-apollo/src/components/AddFeature.tsx
```diff
@@ -110,7 +110,7 @@ export function AddFeature({
       },
       parentFeatureId: sourceFeature._id,
     })
-    changeManager.submit?.(change)
+    await changeManager.submit?.(change)
```
rbuels, line 113: oh man, I would have hoped we had a lint rule that would have caught that

cmdcolin: @typescript-eslint/no-floating-promises may help catch some things like this, it is part of @typescript-eslint/recommended-requiring-type-checking (which needs to be set separately from @typescript-eslint/recommended)

garrettjstevens: @kyostiebi Can you confirm that this change is necessary to fix the issue? I don't see how this would change anything on the backend.

kyostiebi: I just tested that I submitted three times _AddFeatureChange_ -change from UI. Calling _changeManager.submit?.(change)_ without _await_, I got the following error:
[_Nest] 6320  - 06/05/2023, 3:04:46 PM   DEBUG [ParseChangePipe] *** FAILED: MongoServerError: WriteConflict
[Nest] 6320  - 06/05/2023, 3:04:46 PM   DEBUG [ChangesService] *** INSERT DATA EXCEPTION - Start to clean up old temporary documents...
[Nest] 6320  - 06/05/2023, 3:04:46 PM   ERROR [GlobalExceptionsFilter] UnprocessableEntityException: MongoServerError: WriteConflict
[Nest] 6320  - 06/05/2023, 3:04:46 PM   ERROR [GlobalExceptionsFilter] UnprocessableEntityException: MongoServerError: WriteConflict_

With _await_ keyword, all three changes were processed fine.

garrettjstevens:
I wanted to make sure I understood what was going on and what made it so the app-level exception handler missed these exceptions and crashed the app. The main important change ended up being this one:

```diff
diff --git a/packages/apollo-shared/src/Changes/AddFeatureChange.ts b/packages/apollo-shared/src/Changes/AddFeatureChange.ts
--- a/packages/apollo-shared/src/Changes/AddFeatureChange.ts
+++ b/packages/apollo-shared/src/Changes/AddFeatureChange.ts
@@ -155,7 +155,7 @@ export class AddFeatureChange extends FeatureChange {
           })
           const childIds = this.getChildFeatureIds(addedFeature)
           topLevelFeature.allIds.push(addedFeature._id, ...childIds)
-          topLevelFeature.save()
+          await topLevelFeature.save()
         } else {
           const childIds = this.getChildFeatureIds(addedFeature)
           const allIdsV2 = [addedFeature._id, ...childIds]
```

This put the mongoose saving back into the regular flow of the app so the error gets caught and doesn't crash. I also tested with requests not using the UI, and it works well now.

To make sure we don't get any other unexpected crashes, I enabled the `@typescript-eslint/no-floating-promises` lint rule (thanks for the tip @cmdcolin) and fixed the errors that it revealed.

PR #225, opened 2023-06-05
improved details view

### LINKED ISSUE #220: Improve open/close/sizing behavior of details view
Split off from some parts of #216

In `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel/rendering.ts` there is a prop `detailsMinHeight`, a getter `featuresHeight` based on the layout, and a getter `detailsHeight`, based on `detailsMinHeight`, `featuresHeight`, and `height`, which is a getter in `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel/trackHeightMixin.ts`. These, along with `apolloSelectedFeature` in `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel/base.ts`, determine whether the details panel is visible and how big it is.

We could change the behavior of the details panel. We could do something more like Artemis, where the details panel shows a scrollable list of features in the view, and the row corresponding to the selected feature is highlighted in the list. In this case, the details would probably only have two state, open and closed, which would be set by the user, and selecting features wouldn't automatically open the details.

image

If we keep something more like the current behavior, it might be good to increase the overall height of the display when the details are opened instead of opening the details in the existing height, since the current behavior can lead to the selected feature getting covered by the details when they open.

We could also probably use a drag handle between the details and the linear display for easier resizing.

rbuels:
Improves the details view display by:
* removing the "click a feature" message, which improved the scroll bar tidiness
* scrolling the features canvas when the details view pops up so that the selected feature stays visible
* refactoring in preparation for new tabular editing functionality

ref #220

### DIFF
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -2,12 +2,20 @@
   "version": "2.0.0",
   "tasks": [
     {
+      "label": "yarn",
+      "type": "shell",
+      "command": "yarn",
+      "detail": "yarn",
+      "problemMatcher": []
+    },
+    {
       "label": "start shared",
       "type": "npm",
       "script": "start",
       "path": "packages/apollo-shared",
       "problemMatcher": [],
       "detail": "tsc --build --watch",
+      "dependsOn": ["yarn"],
       "isBackground": true
     },
     {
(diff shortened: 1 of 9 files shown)


rbuels:
@garrettjstevens this is probably worth merging now-ish if it looks ok

PR #226, opened 2023-06-09
Add tree tabular view

rbuels:
lots of improvements:

* resizable
* graphical and tabular views can be collapsed
* collapse-all button
* text filter features and highlight matches

image


garrettjstevens (commented):
I've noticed that with Volvox zoomed out to see most features, it can be pretty slow to load all the feature types in the grid. I think it's because it's loading all the autocomplete suggestions for every type in the grid. I think we'll want to only load the suggestions when the autocomplete is opened, similar to this demo: https://mui.com/material-ui/react-autocomplete/#load-on-open

garrettjstevens (commented):
Also, the header of the grid isn't visible in dark mode
localhost_3000__config=http%3A%2F%2Flocalhost%3A9000%2Fjbrowse_config json session=local-LMTm8lbN9C

rbuels:
ok I fixed dark mode and added a cache for the SO terms in the editing component that we can remove once more permanent ontology data store work is in place

PR #229, opened 2023-06-13
Set default assembly in *Delete Assembly* menu

dariober:
In *Delete Assembly* menu display the first assembly availaible instead of showing a blank box.


packages/jbrowse-plugin-apollo/src/components/DeleteAssembly.tsx
```diff
@@ -46,7 +46,17 @@ export function DeleteAssembly({
...
   )
-  const assemblies = useAssemblies(internetAccounts, setErrorMessage)
+
+  const assemblies: AssemblyData[] = useAssemblies(
+    internetAccounts,
+    setErrorMessage,
+  )
+
+  useEffect(() => {
+    if (assemblies.length > 0) {
```
garrettjstevens, line 56: It would be good here to also check if there already is a selected assembly, since we don't need to set one in that case.

garrettjstevens (approved):

PR #231, opened 2023-06-27
Find feature by name/id

shashankbrgowda:
Find feature by name and id functionality

  **Search usage:**
- Enter the feature name to find the feature (no prefix needed for name)
  ex: murB (case insensitive)
image

- To find a feature by its id we need to prefix query with id:
  ex: id:bc2017--bc2017_01552 (case insensitive)
  id:gene:ENSG00000278181
image

**Docker images:**
shashankbrgowda/apollo-collaboration-server-alpha:latest
shashankbrgowda/jbrowse-plugin-apollo-alpha:latest
shashankbrgowda/jbrowse-web-apollo:v2.5.0

### DIFF
--- a/packages/apollo-collaboration-server/src/features/features.controller.ts
+++ b/packages/apollo-collaboration-server/src/features/features.controller.ts
@@ -115,4 +115,26 @@ export class FeaturesController {
     this.logger.debug(`Get all features`)
     return this.featuresService.findAll()
   }
+
+  /**
+   * Fetch feature by its name from featureNameIndex
+   * @param featureName 
+   * @returns a feature object
+   */
+  @Get('name/:name')
+  getFeatureByName(@Param('name') featureName: string) {
+    this.logger.debug(`Get feature by featurename: ${featureName}`)
+    return this.featuresService.findByName(featureName)
+  }
+
+  /**
+   * Fetch feature by its id attribute from featureIdIndex
+   * @param id 
+   * @returns a feature object
+   */
+  @Get('id/:id')
+  getFeatureById(@Param('id') id: string) {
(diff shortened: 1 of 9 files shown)


shashankbrgowda:
I had to make a small change in TextSearchManager relevantAdapters function to include track adapters from the apollo session. While this works for apollo plugin, not sure how this should be incorporated in jbrowse to support both rootmodel and session.

shashankbrgowda/jbrowse-web-apollo:v2.5.0 docker image has this updated code.

```
  relevantAdapters(searchScope: SearchScope) {
    const pm = this.pluginManager
    const { tracks } = pm.rootModel?.session as {
      tracks: AnyConfigurationModel[]
    }

    const { assemblyName } = searchScope

    return [...this.getTrackAdaptersWithAssembly(assemblyName, tracks)]
  }
```

garrettjstevens (commented):
Thanks for the PR @shashankbrgowda. I've opened this PR in the JBrowse 2 repo to make sure we get search adapters in the session tracks: https://github.com/GMOD/jbrowse-components/pull/3767.

I'm still looking at the searching stuff, but I've looked at the xref links and they look like they're working well. I don't think we can include them in the jbrowse_config.json in this repo, though, since that config needs to stay very generic. If you think the xref plugin could be of general utility to others, though, you could consider submitting it to the JBrowse 2 plugin store (see [here](https://github.com/GMOD/jbrowse-plugin-list)), maybe under a name like EBIXrefLinkPlugin since it doesn't depend on Apollo to work.

garrettjstevens:
Implemented in #243 (thanks for the great start on searching!)

PR #232, opened 2023-07-18
ontology manager and ontology storage and retrieval

rbuels:
Adds an OntologyManager that lives on the Apollo session extension, which keeps a list of ontologies with their versions and a FileLocation pointing to their source file. Each ontology in the manager has a view of an OntologyStore that loads the ontology into IndexedDB (only once) and makes it available for querying.

In ontology autocomplete components, removes calls to the external GO API and to the collaboration server.

Also removes the ontology endpoints from the collaboration server.

At some point we are going to want to write some graphical UI to let people manage their ontologies, and the prefixes used to shorten ontology term IRIs.


rbuels:
@garrettjstevens this is ready for review now, could you have a look?

garrettjstevens (approved):
Looks great! I split some stuff out of `session.ts` in #242, so I rebased this to take that into account.

Issue #235, opened 2023-07-18
Add info to feature upload dialog to indicate if an assembly already has features


garrettjstevens:
Add info to feature upload dialog to indicate if an assembly already has features, and also give the option to replace all existing features with the new upload.

garrettjstevens:
Dialog in question is `packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx`

garrettjstevens:
A bit more detail on what this might entail:

When an assembly is selected in the dialog, see if there are any features in that assembly. This probably deserves a new endpoint. I'm thinking this would be something like a Get endpoint in `features.controller.ts` called "getFeatureCount". This endpoint would optionally take assembly ID, refSeq ID, and start and end positions. If given no parameters, it will return a count of all features in the features collection. If given an assembly ID, it will give a count of all the features in that assembly. Same for refSeq, and also same with a refSeq that's limited by start and end. All the logic for this should be in `features.service.ts`.

Then when an assembly is selected in the UI, it calls the "getFeatureCount" endpoint. If there are existing features in that assembly, it shows a message in the dialog that says something like "This assembly already has n features, would you like to delete the existing features before importing new ones?", and then a checkbox they can check if they want to do that.

Then the `AddAssemblyFromFileChange` will need a new option called "deleteExistingFeatures", and then in the `executeOnServer` it will check that option and delete the existing features if necessary.

PR #238, opened 2023-07-22
Check if features already exist on assembly when importing

shashankbrgowda:
While importing features through gff file check if features already exists for the selected assembly, If it exists provide an option to delete existing features before importing or import features without deleting.

Fixes https://github.com/GMOD/Apollo3/issues/235


packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx
```diff
@@ -131,16 +203,46 @@ export function ImportFeatures({
...
+          <DialogContent>
+            <DialogContentText>
+              This assembly already has {featuresCount} features, would you like
+              to delete the existing features before importing new ones?
+            </DialogContentText>
+            <FormControlLabel
+              label={
+                deleteFeatures
+                  ? 'Yes, delete existing features'
+                  : 'No, import without deleting existing features'
```
garrettjstevens, line 226: The label text should not change when the checkbox is clicked, just keep it as 'Yes, delete existing features'. Also, the default should be unchecked.

packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx
```diff
@@ -46,6 +54,64 @@ export function ImportFeatures({
...
   }
 
+  function handleDeleteFeatures(e: React.ChangeEvent<HTMLInputElement>) {
+    setDeleteFeatures(e.target.checked)
+  }
+
+  // fetch and set features count for selected assembly
+  useEffect(() => {
+    if (assemblyId) {
+      updateFeaturesCount()
```
garrettjstevens, line 64: The entire `updateFeaturesCount` function should be inside this `useEffect` block. It's an idiomatic React thing.

packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx
```diff
@@ -46,6 +54,64 @@ export function ImportFeatures({
...
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
garrettjstevens, line 66: Make sure to add all the deps to the `useEffect` deps

packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx
```diff
@@ -46,6 +54,64 @@ export function ImportFeatures({
...
+    })
+
+    if (fetch) {
+      // sumbit might get enabled when we change assembly before loading features count
+      setFeaturesCount(-1)
+      setLoading(true)
+      fetch(uri, {
+        method: 'GET',
+      })
+        .then((res) => res.json())
```
garrettjstevens, line 103: This should be an `async` function and then use `await` instead of chaining `then`s.

packages/jbrowse-plugin-apollo/src/components/ImportFeatures.tsx
```diff
@@ -46,6 +54,64 @@ export function ImportFeatures({
...
+
+  function internetAccount() {
+    const apolloInternetAccounts = internetAccounts.filter(
+      (ia: { type: string }) => ia.type === 'ApolloInternetAccount',
+    ) as ApolloInternetAccountModel[]
+
+    if (apolloInternetAccounts && apolloInternetAccounts.length) {
+      return apolloInternetAccounts[0]
+    }
+  }
```
garrettjstevens, line 77: We need to make sure the internet account is the same one that the selected assembly comes from, since there may be more than one ApolloInternetAccount. Something like this: https://github.com/GMOD/Apollo3/blob/1977709f1a40032cdbe40cf9ed7292a6d939d573/packages/jbrowse-plugin-apollo/src/BackendDrivers/CollaborationServerDriver.ts#L21-L50

garrettjstevens (changes requested):
Looking good, just needs a few changes in the UI component.

garrettjstevens (approved):

PR #240, opened 2023-07-27
don't reload feature data from session snapshot

rbuels:
While working just now I discovered that the assemblies and feature data are being preserved in the session snapshots.

This isn't what we want for obvious reasons.


rbuels:
i'm going to go ahead and merge this and push to staging, I think it might help many of the issues kyosti is seeing

PR #243, opened 2023-08-01
Implement search-by-feature

dariober:
Searches the mongodb *features* collection for matches to a query. A couple of notes:

* This searches all (and only) the documents in the *features* collection. This means that if the loaded assembly(s) do not contain the query but another assembly contains it, you still get hits from the search.

* If a query returns multiple features, the order of the hits in the selection box is in seemingly random order. E.g.:

image

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
+++ b/packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
@@ -17,28 +17,24 @@ export class ApolloTextSearchAdapter
   }
 
   get trackId() {
-    return readConfObject(this.config, 'trackId').uri
+    return readConfObject(this.config, 'trackId')
   }
 
-  get internetAccountPreAuthorization():
-    | { authInfo: { token: string }; internetAccountType: string }
-    | undefined {
-    return readConfObject(this.config, 'baseURL')
-      .internetAccountPreAuthorization
+  get assemblyNames() {
+    return readConfObject(this.config, 'assemblyNames')
   }
 
   async searchFeatures(term: string) {
-    const url = new URL(`features/searchFeatures?term=${term}`, this.baseURL)
+    const assemblies = this.assemblyNames.join(',')
+    const url = new URL('features/searchFeatures', this.baseURL)
+    const searchParams = new URLSearchParams({ assemblies, term })
+    url.search = searchParams.toString()
     const uri = url.toString()
+
     const location: UriLocation = { locationType: 'UriLocation', uri }
-    if (this.internetAccountPreAuthorization) {
-      location.internetAccountPreAuthorization =
-        this.internetAccountPreAuthorization
-    }
     const fetch = getFetcher(location, this.pluginManager)
-    return fetch(uri)
-      .then((res) => res.json())
-      .catch((err) => err)
+    const response = await fetch(uri)
+    return response.json()
   }
 
   mapBaseResult(
@@ -53,18 +49,16 @@ export class ApolloTextSearchAdapter
       (feature) =>
         new BaseResult({
           label: query,
-          displayString: query,
           trackId: this.trackId,
           locString: `${feature.refSeq?.name}:${feature.start}..${feature.end}`,
         }),
     )
   }
 
-  searchIndex(args: BaseTextSearchArgs): Promise<BaseResult[]> {
+  async searchIndex(args: BaseTextSearchArgs): Promise<BaseResult[]> {
     const query = args.queryString
-    return this.searchFeatures(query)
-      .then((features) => this.mapBaseResult(features, query))
-      .catch((err) => err)
+    const features = await this.searchFeatures(query)
+    return this.mapBaseResult(features, query)
   }
 
   // eslint-disable-next-line @typescript-eslint/no-empty-function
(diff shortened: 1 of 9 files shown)


packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+export class ApolloTextSearchAdapter
+  extends BaseAdapter
+  implements BaseTextSearchAdapter
+{
+  get baseURL() {
+    return readConfObject(this.config, 'baseURL').uri
+  }
+
+  get trackId() {
+    return readConfObject(this.config, 'trackId').uri
```
garrettjstevens, line 20: No `uri` property here
```suggestion
    return readConfObject(this.config, 'trackId')
```

packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+  }
+
+  get internetAccountPreAuthorization():
+    | { authInfo: { token: string }; internetAccountType: string }
+    | undefined {
+    return readConfObject(this.config, 'baseURL')
+      .internetAccountPreAuthorization
+  }
```
garrettjstevens, line 28: I don't think this does anything and can probably be safely removed.

packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  async searchFeatures(term: string) {
+    const url = new URL(`features/searchFeatures?term=${term}`, this.baseURL)
```
garrettjstevens, line 31: It's probably good to use `URLSearchParams` to create this so it can e.g. handle conversion of characters (like spaces).
```suggestion
    const url = new URL('features/searchFeatures', this.baseURL)
    const searchParams = new URLSearchParams({ term })
    url.search = searchParams.toString()
term=${term}`, this.baseURL)
```

packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+    const uri = url.toString()
+    const location: UriLocation = { locationType: 'UriLocation', uri }
+    if (this.internetAccountPreAuthorization) {
+      location.internetAccountPreAuthorization =
+        this.internetAccountPreAuthorization
+    }
+    const fetch = getFetcher(location, this.pluginManager)
+    return fetch(uri)
+      .then((res) => res.json())
+      .catch((err) => err)
```
garrettjstevens, line 41: This should be in async-await style for consistency with existing code.
```suggestion
    const response = await fetch(uri)
    return response.json()
```

packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+      start: number
+      end: number
+    }[],
+    query: string,
+  ) {
+    return features.map(
+      (feature) =>
+        new BaseResult({
+          label: query,
+          displayString: query,
```
garrettjstevens, line 56: I think we can leave `displayString` out, since it's the same as `label`

packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
```diff
@@ -0,0 +1,72 @@
...
+          locString: `${feature.refSeq?.name}:${feature.start}..${feature.end}`,
+        }),
+    )
+  }
+
+  searchIndex(args: BaseTextSearchArgs): Promise<BaseResult[]> {
+    const query = args.queryString
+    return this.searchFeatures(query)
+      .then((features) => this.mapBaseResult(features, query))
+      .catch((err) => err)
```
garrettjstevens, line 67: This should be in async-await style for consistency with existing code.
```suggestion
  async searchIndex(args: BaseTextSearchArgs): Promise<BaseResult[]> {
    console.log({ args })
    const query = args.queryString
    const features = await this.searchFeatures(query)
    return this.mapBaseResult(features, query)
```

garrettjstevens (changes requested):
One fix I noticed this needs is that the text index needs to be added to `feature.schema.ts`, e.g. `FeatureSchema.index({ '$**': 'text' })`.

I think I know a way to keep this from returning results from multiple assemblies. First of all you can check the assemblies that need to be searched by accessing the `assemblyNames` config of the adapter, e.g.:
```js
get assemblyNames() {
  return readConfObject(this.config, 'assemblyNames')
}
```
Then we need to add another term to our URL params, so a request will look something like this: `http://localhost:3999/features/searchFeatures?assemblies=643da4cad6cf56a77eb695d7,644c4e0266ee971b223d952e&term=exonerate` (usually there will just be one assembly, but we want to be able to search multiple in the future).

Then when searching in MongoDB, we should limit results to those that match the assemblies.

PR #244, opened 2023-08-01
Add fulltext indexing to ontology store

rbuels:
Adds full-text indexing to the ontology store and uses it for autocomplete.

Good results, pretty fast and effective.

This also has a Jest upgrade in it that was part of my attempts to make inline snapshots work again. Jest inline snapshots no longer work since we upgraded to prettier 3 (probably related to https://github.com/prettier/eslint-plugin-prettier/issues/562, seems likely that a fix will be coming soon)


rbuels:
@garrettjstevens ok I think this is good to go

garrettjstevens:
Wow, that does make it a lot faster. I have noticed one thing, though, that can make the search results a bit unintuitive. For example, here's a search for "calcium" before this branch:
image

And in this branch:
image

It looks like the top terms that pop up in this branch do have "calcium" in their description, but not the label, so it's maybe not clear to the user why they show up. Maybe the indexing should rank search results with the term in the label above those with the term in the description.

rbuels:
@garrettjstevens i think this is pretty good, have another look and merge if you like

PR #281, opened 2023-09-14
Add sequence to exported GFF3

kyostiebi:
FASTA is now included in exported GFF3. Please, merge this only after [Bug_251_Download_not_working](https://github.com/GMOD/Apollo3/tree/Bug_251_Download_not_working) has been merged.

### DIFF
--- a/packages/apollo-collaboration-server/src/features/features.service.ts
+++ b/packages/apollo-collaboration-server/src/features/features.service.ts
@@ -241,9 +241,14 @@ export class FeaturesService {
         .find({ refSeq: refSeqDoc.id })
         .sort({ n: 1 })
         .cursor()) {
-        printFasta ? sequenceStream.push('##FASTA\n') : null
-        sequenceStream.push(`>${refSeqDoc.name} ${refSeqDoc.description}\n`)
-        // eslint-disable-next-line unicorn/no-array-push-push
+        if (printFasta) {
+          sequenceStream.push('##FASTA\n')
+          refSeqDoc.description
+            ? sequenceStream.push(
+                `>${refSeqDoc.name} ${refSeqDoc.description}\n`,
+              )
+            : sequenceStream.push(`>${refSeqDoc.name}\n`)
+        }
         sequenceStream.push(`${this.splitStringIntoChunks(doc.sequence, 60)}\n`)
         printFasta = false
       }
--- a/packages/apollo-collaboration-server/src/features/features.service.ts
+++ b/packages/apollo-collaboration-server/src/features/features.service.ts
@@ -11,6 +11,8 @@ import {
   Feature,
   FeatureDocument,
   RefSeq,
+  RefSeqChunk,
+  RefSeqChunkDocument,
   RefSeqDocument,
 } from 'apollo-schemas'
 import { GetFeaturesOperation } from 'apollo-shared'
@@ -19,6 +21,7 @@ import StreamConcat from 'stream-concat'
 
 import { FeatureRangeSearchDto } from '../entity/gff3Object.dto'
 import { OperationsService } from '../operations/operations.service'
+import { RefSeqChunksService } from '../refSeqChunks/refSeqChunks.service'
 import { FeatureCountRequest } from './dto/feature.dto'
 
 function makeGFF3Feature(
@@ -131,6 +134,7 @@ function makeGFF3Feature(
 export class FeaturesService {
   constructor(
     private readonly operationsService: OperationsService,
+    private readonly refSeqChunkService: RefSeqChunksService,
     @InjectModel(Feature.name)
     private readonly featureModel: Model<FeatureDocument>,
     @InjectModel(Assembly.name)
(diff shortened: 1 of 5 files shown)


packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -196,19 +212,38 @@ export class FeaturesService {
...
+
     headerStream.push('##gff-version 3\n')
     for (const refSeqDoc of refSeqs) {
       headerStream.push(
         `##sequence-region ${refSeqDoc.name} 1 ${refSeqDoc.length}\n`,
       )
+      for await (const doc of this.refSeqChunksModel
+        .find({ refSeq: refSeqDoc.id })
+        .sort({ n: 1 })
+        .cursor()) {
```
garrettjstevens, line 235: Each ref seq might contain multiple ref seq chunks, so you'll need to take that into account. For example, if I set `CHUNK_SIZE` in the env file to 20,000 and then import a Volvox assembly, the exported FASTA looks like this:
```
##FASTA
>ctgA 
cattgttgcggagttgaacaACGGCATTAGGAACACTTCCGTCTCtcacttttatacgat
...
gtaagcgataggcaaactgcggttattcgcgtagaccgcttgatgatgtaaaataacagt
##FASTA
>ctgA 
cattgttgcggagttgaacaACGGCATTAGGAACACTTCCGTCTCtcacttttatacgat
...
cgaacaccgtgatccattga
>ctgA
ggcgtggaggcatagggccggagtgttcccgtaccaattacgagctaagaccgcctgtcc
...
aaaaacgaattacgtgtatagagattatgtcactccgtcac
>ctgB
ACATGCTAGCTACGTGCATGCTCGACATGCATCATCAGCCTGATGCTGATACATGCTAGC
...
```
So there are multiple ctgAs, and ##FASTA shows up twice somehow.

packages/apollo-collaboration-server/src/features/features.service.ts
```diff
@@ -196,19 +212,38 @@ export class FeaturesService {
...
+        printFasta ? sequenceStream.push('##FASTA\n') : null
+        sequenceStream.push(`>${refSeqDoc.name} ${refSeqDoc.description}\n`)
```
garrettjstevens, line 237: If there is no description, this should not print a space after the name so there aren't any trailing spaces.

garrettjstevens (changes requested):

packages/apollo-collaboration-server/src/features/features.service.ts
kyostiebi: Fixed

packages/apollo-collaboration-server/src/features/features.service.ts
kyostiebi: Now, there is just one time **##FASTA** and RefSeqs name and description even sequence was saved in several chunks in Mongo. However, due to multiple chunks per sequence, sequence lines are not always same length i.e. output may look like this:
_...
tgggtagagcgccgatagctaattctcgagtggaactgcggaatgtgtcatatgaggtag
gcacggccctggcgtggcggcggtacgcgtacaaacctccactgcttggggaataaggtg
cgaacaccgtgatccattga
ggcgtggaggcatagggccggagtgttcccgtaccaattacgagctaagaccgcctgtcc
agccctggtgattacgtagtaagctcggtacaagcctgtaattagcttccgggtggcgga
cattcgttcatattgccagtcacggcagaggccgcgctagctaggccagccccgatgaaa
..._

I did not find any spec that each sequence line should be same length so I guess this is not an issue?

garrettjstevens: They do need to be the same length (except for the last line), not necessarily because of the spec, but because most parsers/indexers that people use on this output will expect them the be the same length.

kyostiebi: Now added the following property
# Sequence line lenght in exported GFF3 file
SEQ_LINE_LEN=90

All lines (except the last one) are now this long

garrettjstevens:
Has this been tested on exporting very large data? Looking at the code, it looks like it's loading all the sequences into memory, which we want to avoid because we want to keep server memory usage low, and it will be slower. I think for the `sequenceStream` it will need to use a `pipeline` like the `featureStream` does. Just as an outline, that might look like this:

```js
pipeline(
  this.refSeqChunksModel
        .find({ refSeq: refSeqDocs }) // get chunks for all refSeqs in a single query
        .sort({ n: 1 }) // make sure the sort accounts for having all refSeqs
        .cursor(),
  new Transform({
    construct(callback) {
      this.printFasta = true;
      this.printSeqName = true;
      this.remainingLastLine = ''
      callback();
    },
    transform(chunk, encoding, callback) {
      // basically the same logic that's already there, but using `this.push` instead of `sequenceStream.push`
      callback();
    },
    flush(callback) {
      if (this.remainingLastLine) this.push(remainingLastLine)
    },
  }),
  (error) => {
    if (error) {
      this.logger.error('GFF3 export failed')
      this.logger.error(error)
    }
  },
);
```

The node docs on streams might be helpful here: https://nodejs.org/docs/latest-v18.x/api/stream.html

garrettjstevens (approved):

PR #289, opened 2023-09-28
Add Apollo assembly and feature importing status to JBrowse 2 jobs widget

### LINKED ISSUE #234: Add Apollo assembly and feature importing status to JBrowse 2 jobs widget
Show upload and parsing status of assembly and/or features in the JBrowse job manager so that users can easily see the status of their data uploading.

The first step is to see if jobs can be added to the jobs widget without needing to use the JBrowse job manager. If it turns out the jobs would benefit from being queued (e.g. they are using more local system resources than expected), then see if we can use the job manager.

carolinebridge-oicr:
Resolves #234

Dependent on JBrowse core changes: https://github.com/GMOD/jbrowse-components/pull/3951 (merged), changes will be in v2.7.0

## summary of changes
- Adds a new ApolloJobModel to update the jobs widget UI where appropriate
- Adds a safeguard to the add assembly UI to prevent a user from uploading an assembly with an existing name (an operation that will always fail and takes some time if the file is large)
- Utilizes the new jobs manager to update the widget UI when a user attempts to upload a new assembly or upload new features for an assembly


carolinebridge-oicr:
^ tests are failing because `WidgetType 'JobsListWidget' not found`, which is included in the jbrowse-components PR

packages/jbrowse-plugin-apollo/src/ChangeManager.ts
```diff
@@ -22,22 +24,41 @@ export class ChangeManager {
...
   async submit(change: Change, opts: SubmitOpts = {}) {
-    const { addToRecents = true, submitToBackend = true } = opts
+    const {
+      addToRecents = true,
+      submitToBackend = true,
+      updateJobsManager = false,
+    } = opts
     // pre-validate
     const session = getSession(this.dataStore)
+    const { jobsManager } = session
```
garrettjstevens, line 34: Getting the session vs Apollo session right can be tricky, but I think this should give you the right typings:

```suggestion
    const { notify } = getSession(this.dataStore)
    const { jobsManager } = getSession(
      this.dataStore,
    ) as unknown as ApolloSessionModel
```

garrettjstevens (approved):

garrettjstevens:
After merging main, it looks like the tests aren't passing because the "Features are being added" message that "importFeatures" is looking for does not exist after this change. I tried to update the test, but didn't have any luck. @dariober, would you be able to take a look at this? Instead of checking for the previous message, I think you can now tell features have been added when the "Completed jobs" section has two entries that both say "All operations successful".

image

garrettjstevens:
@dariober never mind, I was able to figure out the tests :)

Issue #286, opened 2023-09-21
"Loading" and "No data" messages


garrettjstevens:
In the graphical display and the tabular display, it would be good to have a "Loading" message while the data is being loaded so the user knows that something is happening.

When the loading is finished, if there is no data to show in the displayed area, the tabular display should also show a "No data in this region" or some sort of other message.

May need to add a loading status flag to the client data store.

garrettjstevens:
Possible implementation for loading message:
- Add "loadingRegions" as a `Region[]` to volatile section of `packages/jbrowse-plugin-apollo/src/session/ClientDataStore.ts`
- In `loadFeatures` in that same file, add currently loading region to "loadingRegions" at the beginning of the action and remove it at the end
- In `packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/LinearApolloDisplay.tsx`, check the currently loading regions (via `session.apolloDataStore.loadingRegions`) and the displayed regions (via `model.displayedRegions`), and if any of them overlap, add a "Loading..." message to the component
  - The loading message can be in the corner of the track and overlap the regular track area, since it may still be loading new data when there is already data on the screen

I'll have to think more about the "No data" message

PR #303, opened 2023-10-13
Adds "loading" and "no data" messages to graphical and tabular displays when relevant

carolinebridge-oicr:
resolves https://github.com/GMOD/Apollo3/issues/286

## summary of changes
- adds loading bar to graphical and tabular displays when data is loading and glyphs are being drawn, these messages are dismissed based on when the last loaded element has been drawn in the autorun
- ⚠️ adds filtering to the "seenFeatures" for the tabular display such that what is written in the table is the same as what can be seen in the graphical display -- **needs review**
- adds a "no data" message when there is no data to be displayed in the tabular view -- that is, that there is nothing drawn on the graphical display

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/TabularEditor/HybridGrid/HybridGrid.tsx
+++ b/packages/jbrowse-plugin-apollo/src/TabularEditor/HybridGrid/HybridGrid.tsx
@@ -1,6 +1,7 @@
 import { Menu, MenuItem } from '@jbrowse/core/ui'
 import { getSession } from '@jbrowse/core/util'
-import { useTheme } from '@mui/material'
+import LoadingEllipses from '@jbrowse/core/ui/LoadingEllipses'
+import { Alert, useTheme } from '@mui/material'
 import { observer } from 'mobx-react'
 import React, { useEffect, useMemo, useRef, useState } from 'react'
 import { makeStyles } from 'tss-react/mui'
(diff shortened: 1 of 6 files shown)


packages/jbrowse-plugin-apollo/src/TabularEditor/HybridGrid/HybridGrid.tsx
```diff
@@ -38,14 +59,24 @@ const HybridGrid = observer(function HybridGrid({
...
-
+  const { loadingRegions } = getSession(model).apolloDataStore
   const { filterText } = tabularEditor
 
+  // filters seenFeatures by features only directly observed by the user
+  const [visibleRegion] = lgv.getSelectedRegions()
+  const visibleFeatures = [...seenFeatures.entries()].filter((entry) => {
+    const [, feature] = entry
+    return (
+      feature.start > visibleRegion.start && feature.end < visibleRegion.end
```
carolinebridge-oicr, line 76: @garrettjstevens 

Adds filtering to the seenFeatures property such that what's displayed on the table is only what is being observed in the graph (previously, it showed everything that had been loaded). "No data" message will display when this value is empty.

Not sure if this is correcting broken behaviour or if it should be kept as showing everything that's currently been loaded. Might be a problem with `model.displayedRegions` ?

carolinebridge-oicr (commented):
@garrettjstevens 

Let me know if the UI aligns with the styling you had in mind. Currently both messages are `zIndex: 10` hovering in the top right corner of the graphical and tabular views.

packages/jbrowse-plugin-apollo/src/TabularEditor/HybridGrid/HybridGrid.tsx
carolinebridge-oicr: Re: styling, for the table it might be better to have it centred in the table like a data grid MUI component, subjective though

PR #312, opened 2023-11-01
Initial check implementation

### LINKED ISSUE #308: Add check collection and query endpoints
The branch [add_check_initial](https://github.com/GMOD/Apollo3/tree/add_check_initial) has a check result schema defined, but it's not used yet. On a new branch based off of this branch:
- Register the check result schema in the collaboration server backend
- Add endpoints in checks.controller.ts to:
  - query all checks in a range
  - query checks by associated feature's ID

### LINKED ISSUE #309: Generate check results for locally opened files
In the branch [add_check_initial](https://github.com/GMOD/Apollo3/tree/add_check_initial):
- Make a new directory in `packages/apollo-shared/src` called "Checks"
- Make an abstract base class "Check" in that directory for checks
- I think the base class for now just needs one method with the signature:
`async checkFeature(feature: AnnotationFeatureSnapshot, sequence: string): CheckResultSnapshot`
- Import that base class in `packages/jbrowse-plugin-apollo/src/BackendDrivers/InMemoryFileDriver.ts` and create a check using it (can be just a fake check for now, like in `packages/apollo-collaboration-server/src/checks/checks.service.ts`)
- In that same file in the `for (const seqLine of featuresAndSequences) {` loop
  - Call the check on the feature generated by `const feature = createFeature(seqLine)`
  - Make list of non-empty check results
  - After all checks have been made, add them to the client store with `apolloDataStore.addCheckResults`

You can check if this works by using the menu item "Open local GFF3 file" to open a small file and then running `getSnapshot(JBrowseSession.apolloDataStore.checkResults)` in the browser console to see if there are any results there.

### LINKED ISSUE #314: Use Check class to run checks
This can be done on the same branch as #312 after the comments there have been addressed.

- Import the `Check` base class and create a check in `checks.service.ts` (a fake check just like the one in `OpenLocalFile.tsx` is fine for now)
- Call the new class's `checkFeature` method in `checks.service.ts`'s `checkFeature`
  - You'll have to look up the sequence for the feature to pass it to the check

You might want to create a temporary testing endpoint that takes a feature ID, looks up the feature, runs `checkFeature` on it, and returns the result.

FYI the next step will be to start registering individual checks that will all be called in `check.service.ts`'s `checkFeature`.

garrettjstevens:
Resolves #308
Resolves #309
Resolves #314


packages/apollo-collaboration-server/src/checks/checks.service.ts
```diff
@@ -0,0 +1,80 @@
...
+  }
+
+  /**
+   * Get all possible checkReports for given featureId
+   * @param id - featureId
+   * @returns - an array of checkReport -documents
+   */
+  async findByFeatureId(id: string) {
+    const checkReports = await this.checkReportModel.find({ ids: id }).exec()
+    return checkReports
```
garrettjstevens, line 48: @kyostiebi Instead of `await`ing something and then immediately returning the result, you should return the result directly instead.

```suggestion
    return this.checkReportModel.find({ ids: id }).exec()
```

packages/apollo-collaboration-server/src/checks/checks.service.ts
```diff
@@ -0,0 +1,80 @@
...
+        typeName: 'GetFeaturesOperation',
+        refSeq: searchDto.refSeq,
+        start: searchDto.start,
+        end: searchDto.end,
+      })
+    const allIdsArray: string[] = featureDocs.flatMap((doc) => doc.allIds)
+    const checkReports = await this.checkReportModel
+      .find({ ids: { $in: allIdsArray } })
+      .exec()
+    return checkReports
```
garrettjstevens, line 68: @kyostiebi With the `checkResultModel`, you can query the range directly without using feature IDs since it has `refSeq`, `start`, and `end`.

packages/apollo-schemas/src/checkReport.schema.ts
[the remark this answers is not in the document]
kyostiebi: Done

packages/apollo-collaboration-server/src/checks/checks.module.ts
```diff
@@ -0,0 +1,29 @@
+import { Module, forwardRef } from '@nestjs/common'
+import { MongooseModule, getConnectionToken } from '@nestjs/mongoose'
+import { CheckReport, CheckReportSchema } from 'apollo-schemas'
```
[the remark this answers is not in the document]
kyostiebi: Done

packages/apollo-collaboration-server/src/checks/checks.service.ts
kyostiebi: Done

packages/apollo-collaboration-server/src/checks/checks.service.ts
kyostiebi: Done

PR #321, opened 2023-11-15
Render checks in track

### LINKED ISSUE #307: Render check results on the LinearApolloDisplay
Using the branch [add_check_initial](https://github.com/GMOD/Apollo3/tree/add_check_initial), render the contents of the checkResults (stored in the client data store) in the LinearApolloDisplay. As a first pass, we could do some small symbol at the beginning of the glyph with a tooltip with check info.

garrettjstevens:
Fixes #307 

image


garrettjstevens:
@dariober can you take a look at the failing test here?

PR #322, opened 2023-11-15
Add checks to assembly

### LINKED ISSUE #317: Add check management UI
Using the branch [add_checks_to_assembly](https://github.com/GMOD/Apollo3/tree/add_checks_to_assembly):

- Add a version to both `check.schema.ts` and `Check.ts`
- When the server starts, it goes through all the checks registered in `checkRegistry` and:
  - If the check doesn't exist in the checks collection, add it
  - If the check does exist and the version is the same, do nothing
  - If the check does exist and the version is different, update the `updatedAt` field 
- Add a menu item in the Apollo menu called "Manage checks"
- Create a dialog similar to "ManageUsers.tsx" or "ViewChangelog.tsx" that shows:
  - All checks that have been added to the check collection
    - This is where a check can be set to "default"
  - All checks assigned to each assembly
    - The `assembly.schema.ts` has been updated in this branch to allow it to reference checks
  - Allows you to add/remove checks from an assembly, and change whether checks are default
- Also, wherever in the code an assembly is created, add any checks marked as default in the check collection to it.

If you need a check in the check collection to start testing with, you can use:

```js
use('apolloDb');
db.getCollection('checks').insertOne({
  name: 'CDSCheck',
  version: 1,
  default: true,
  createdAt: new Date(),
  updatedAt: new Date(),
});
```

garrettjstevens:
Resolves #317 

image


garrettjstevens:
@dariober @kyostiebi I think the issue Dario saw with getting an error when submitting checks should be resolved now.

Issue #349, opened 2024-01-23
Color CDS by reading frame


garrettjstevens:
As you can see in the screenshot below, Apollo 2 gave each reading frame a color and then colored the CDS to match its frame. I'd like to see how this looks in Apollo 3. So the idea is:

- Color each translation rows (e.g the pink, green, and blue in the screenshot)
- Color the CDS the color of its associated row (i.e. the row that highlights when it is hovered over)

Feel free to use the same color scheme or pick a new one.

![image](https://github.com/GMOD/Apollo3/assets/25592344/868ed6e3-d2ec-4b96-a377-eaf9dd01f2e3)

This should be put in as a separate PR that builds on https://github.com/GMOD/Apollo3/pull/341.

PR #353, opened 2024-01-25
Color CDS by reading frame

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/349

image


garrettjstevens:
There's some potential we could coordinate with JBrowse core to unify coloring, etc. of the reading frames. See https://github.com/GMOD/jbrowse-components/pull/4185.

PR #354, opened 2024-02-01
First Apollo CLI commands: config, login, logout, and status

dariober:
The `login` subcommand reads the config file for a given profile, prompts the user to login as appropriate (if e.g. accessType is google open the web browser), finally stores the access token in the config file under the selected profile.


dariober:
I'm not sure whether the linter errors can be avoided or they should be ignored.

packages/apollo-cli/src/Config.ts
```diff
@@ -0,0 +1,184 @@
...
+  username?: string
+  password?: string
+}
+
+interface Profile {
+  address?: string
+  accessType?: string
+  accessToken?: string
+  rootCredentials?: RootCredentials
+}
```
garrettjstevens, line 18: Could you change the type for the Profile to be this:

```ts
interface BaseProfile {
  address: string
  accessType: 'google' | 'microsoft' | 'guest'
  token: string
}

interface RootProfile extends Omit<BaseProfile, 'accessType'> {
  accessType: 'root'
  rootCredentials: {
    username: string
    password: string
  }
}

export type Profile = BaseProfile | RootProfile
```

This way we can depend on "address" and "accessType" not being undefined, and "rootCredential" being defined only if "accessType" is "root".

packages/apollo-cli/src/Config.ts
```diff
@@ -0,0 +1,184 @@
...
+const profileSchema = Joi.object({
+  address: Joi.string().custom(validateAddress),
+  accessType: Joi.string().valid('google', 'microsoft', 'root'),
+  accessToken: Joi.string(),
+  rootCredentials: {
+    username: Joi.string().allow(null, ''),
+    password: Joi.string().allow(null, ''),
+  },
+})
+const configSchema = Joi.object().pattern(/.*/, profileSchema)
```
garrettjstevens, line 67: This is related to the above comment on the Profile type, I think this schema captures the same thing we're trying to describe in the updated type.

```suggestion
const profileSchema = Joi.object({
  address: Joi.string()
    .uri({ scheme: /https?/ })
    .required(),
  accessType: Joi.string()
    .valid('google', 'microsoft', 'root', 'guest')
    .required(),
  accessToken: Joi.string(),
  rootCredentials: Joi.object({
    username: Joi.string().required(),
    password: Joi.string().required(),
  }).when('accessType', {
    is: Joi.string().valid('root'),
    // eslint-disable-next-line unicorn/no-thenable
    then: Joi.required(),
    otherwise: Joi.forbidden(),
  }),
})
const configSchema = Joi.object().pattern(Joi.string(), profileSchema)
```

garrettjstevens (changes requested):

packages/apollo-cli/src/Config.ts
```diff
@@ -0,0 +1,189 @@
...
+import Joi from 'joi'
+import YAML from 'yaml'
+
+import { ConfigError } from './utils.ts'
+
+interface BaseProfile {
+  address: string
+  accessType: 'google' | 'microsoft' | 'guest'
+  accessToken: string
+  token: string
```
garrettjstevens, line 12: `accessToken` should be optional, and `token` is unneeded

garrettjstevens (changes requested):
I pushed a few cleanup changes. Here is a checklist of things I think still need to be done for these commands:

- [x] Fix bug where running `apollo config` and setting up a new profile of any other type than "root" results in an error. (I think this is new as of the last commit.)
- [x] Fix bug where `apollo config` fails if the `config.configDir` directory doesn't exist
- These items apply to all of `apollo login`, `apollo logout`, and `apollo status`:
  - [x] If there is no config directory, no config file, or an empty config file, all three of these commands should exit with an error message that is something like: "The Apollo CLI has not been configured yet. Please run `apollo config` to configure it."
  - [ ] If e.g. `--profile abcd` is provided to any of these three commands and that profile does not exist, they should exit with a message saying that the profile could not be found and to run `yarn config` to set it up.
- [x] Make `yarn logout` remove `accessToken` from whatever profile is specfied in the config file
- [x] Make `yarn status` print "Logged in" if there is an `accessToken` for the specified profile in the config file, otherwise it should print "Logged out"
- [x] If there are not any profiles yet, skip the "Select profile to edit or create a new one" question
- [x] Make the `login` and `logout` commands not mention Keycloak
- [ ] Fix bug where `"default.rootCredentials.password" must be a string` is displayed if "guest" is selected during `yarn config`
- [ ] Fix bug where `apollo login` while using a "guest" type profile does not save the `accessToken`. Currently it opens a browser window, but it does not need to, it should get the token in the same way a "root" login does
- [ ] Make sure all commands print a sensible error message if the config file is not valid YAML or doesn't pass validation

garrettjstevens:
Make sure tests work if config dir doesn't exist yet
```diff
diff --git a/packages/apollo-cli/src/test/fixtures.ts b/packages/apollo-cli/src/test/fixtures.ts
index 30aa7a4d..f7117875 100644
--- a/packages/apollo-cli/src/test/fixtures.ts
+++ b/packages/apollo-cli/src/test/fixtures.ts
@@ -4,10 +4,8 @@ import path from 'node:path'
 
 export const TEST_DATA_DIR = path.resolve('test_data')
 export const VERBOSE = false
-export const CONFIG_FILE = path.join(
-  os.homedir(),
-  '/.config/apollo-cli/config.yaml',
-)
+export const CONFIG_DIR = path.join(os.homedir(), '.config', 'apollo-cli')
+export const CONFIG_FILE = path.join(CONFIG_DIR, 'config.yaml')
 const CONFIG_BAK = path.join(TEST_DATA_DIR, 'original.config.yaml.bak')
 
 function renameFile(src: string, dest: string, verbose = true) {
@@ -28,6 +26,7 @@ function renameFile(src: string, dest: string, verbose = true) {
 }
 
 export function copyFile(src: string, dest: string, verbose: boolean) {
+  fs.mkdirSync(path.dirname(dest), { recursive: true })
   fs.copyFileSync(src, dest)
   if (verbose) {
     const msg = `cp ${src} ${dest}`

```

garrettjstevens (approved):

PR #356, opened 2024-02-15
Cli queries

dariober:
Add query commands `assemblies`, `refSeqs`, `features` with basic functionality. There is also some refactoring of the `Config` and `BaseCommand` to avoid code duplication.

### DIFF
--- a/.husky/_/.gitignore
+++ b/.husky/_/.gitignore
@@ -1,2 +1,1 @@
 *
-!.gitignore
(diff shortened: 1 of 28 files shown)


ihh:
One suggestion I had was to keep everything lower case eg “refseqs” not
“refSeqs”, since upper case is generally viewed as a complication in
command line UIs.

garrettjstevens:
These changes were included in the merge of https://github.com/GMOD/Apollo3/pull/379

Issue #357, opened 2024-02-20
Add ability to get parent ID in feature from features endpoint


garrettjstevens:
From Havana, when they process the changes, they need the parent ID of a changed feature. This could possibly be added in `packages/apollo-common/src/FeatureChange.ts`.

@shashankbrgowda, could you expand a bit more on the use case for this?

shashankbrgowda:
We need to add parent id attribute when we process the change (Coordinate change / Attribute change etc.) 
or retain the "Parent" field in the attributes which is present in gff, so that when we get feature its already available.

```
1	ensembl_havana	exon	177280412	177282422	.	+	.	Parent=transcript:ENST00000361539;Name=ENSE00001166070;constitutive=0;ensembl_end_phase=-1;ensembl_phase=2;exon_id=ENSE00001166070;rank=8;version=9
```

garrettjstevens:
We're going to go the route of obtaining the parent ID from the endpoint call instead.

In `packages/apollo-collaboration-server/src/features/features.controller.ts` `@Get(':featureid')` endpoint, add the ability to have a URL parameter that requests that the parent ID be returned as well (e.g. `/features/624ab4c0f8ac0187ed22b563?parentId=true`).

In `packages/apollo-collaboration-server/src/features/features.service.ts`, first the parent feature is queried from the database and then if the requested feature is a child feature, it returns that sub-document of the parent feature. We will add an option if parentId is true and the requests feature is a child feature, then the parent ID is added to childFeature.attributes before returning it.

garrettjstevens:
After discussion, the we're going to try a different solution and see how the two compare. The new solution we are going to try is having a URL parameter called `topLevel`, used like: `features/624ab4c0f8ac0187ed22b563?topLevel=true`. Then instead of just returning the feature and its children, it will return the top-level parent feature of this feature and all its children.

PR #361, opened 2024-02-26
Add ability to get parent ID in feature from features endpoint

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/357


shashankbrgowda:
Fixed by https://github.com/GMOD/Apollo3/pull/456

PR #371, opened 2024-03-21
Get config.json from collaboration server

kyostiebi:
Apollo tracks are now retrieved from Mongo and saved in config.json


kyostiebi:
I just pushed an updated version

Issue #373, opened 2024-03-27
Data setup documentation


garrettjstevens:
Now that we have deployment documentation in #372, we need to document (and also in some cases do some development) how users add assemblies, annotations, and evidence tracks. We'll need to document this process for users who want to do it via the CLI and for those who want to use the UI.

For setting up assemblies and annotations in the UI, we have a pretty established workflow, so mostly what's left there is to document it. For setting up assemblies and annotations from the CLI, we have commands that do that as well. It might be good to flesh out the "description" of the various flags and args in the CLI command classes, as well as adding some "examples" to the command classes (as seen [here](https://oclif.io/docs/commands#other-command-options)).

Adding evidence tracks is where we need to finalize some more things. One workflow that @kyostiebi has started working on is selecting a track that was added as a session track to JBrowse in the UI and then copying that to the config.json on the collaboration server. This is fine if the track points to a publicly hosted URL, but if the user wants to upload an evidence track file to the server, how do we deal with that?

Similarly, how do we add evidence tracks from the CLI? We can, for example, create a config.json and use the JBrowse CLI to add the evidence tracks to it, and then copy over the track configs to the collaboration server. But how do we make sure the files are in the right place on the server? And if the site admin has to use the JBrowse CLI and the Apollo CLI on the remote server, should that be part of one of the Docker containers in the existing compose.yml, or should it be something separate?

@kyostiebi @dariober @shashankbrgowda feel free to approach this and split it up however you like, and let me know if you need anything from me and I'll see it when I'm back next week.

garrettjstevens:
@kyostiebi @dariober @shashankbrgowda any updates from your discussion last week?

shashankbrgowda:
> @kyostiebi @dariober @shashankbrgowda any updates from your discussion last week?

Documented steps to upload assembly, import features and delete an assembly. https://github.com/GMOD/Apollo3/pull/376

dariober:
I prepare [this](https://github.com/GMOD/Apollo3/blob/cli_data_setup_docs/docs/data_setup/cli.md) but the PR is failing https://github.com/GMOD/Apollo3/pull/374

PR #374, opened 2024-03-28
First commit of cli docs for data setup

dariober:
First pass at addressing issue #373 with respect to cli usage for data set setup. 

It assumes Apollo is up and running but with nothing in it. So the docs start by setting up a user account. Then it describes how to add data except evidence tracks. It doesn't cover how to edit features or query the database (I guess that would be a separate document?). See if this goes in the right direction...!


garrettjstevens:
Superseded by docs in `packages/website`

PR #379, opened 2024-04-08
Add more CLI functionality

dariober:
Several additions to the CLI


garrettjstevens:
Hi @dariober, I couldn't ever get the FormData to stream properly, so I tried another approach, creating a new endpoint to stream files to without them being in FormData. The branch is here:
https://github.com/GMOD/Apollo3/tree/cli-misc-gjs

Could you try out this branch and see if it works on your end?

Issue #391, opened 2024-05-15
cli via docker does not edit the local config file


dariober:
When running the cli via docker we need to check were it reads and writes the config file. Currently, this:

```
docker run apollo-cli config address http://localhost:1234

yarn run v1.22.19
warning package.json: No license field
$ tsx bin/dev.js config address http://localhost:1234
"default.accessType" is required
Done in 0.75s.
```

Completes fine, but the local configuration file is not changed

garrettjstevens:
We'll need to use a bind mount or a volume, probably. With a [bind mount](https://docs.docker.com/storage/bind-mounts/), I think it would look like `-v "~/.config:/path/to/container/.config`, where we'd have to figure out where the `.config` directory for the default user is in the container. This would make it so any changes in `.config` in the container would also happen in `.config`.

I think [volumes](https://docs.docker.com/storage/volumes/) are supposed to be a safer way of doing something similar, but I'm not as familiar as them. Basically you create a volume with a name, and then you can mount that volume every time the container is run and persist data that way without having to make any potentially conflicting changes to your local .config directory.

dariober:
This Dockerfile, which is suspiciously simple, works...:

```
FROM node:18-alpine3.19
WORKDIR /app
RUN npm install -g @apollo-annotation/apollo-cli
```

... but usage seems a bit laborious.

First build a container named `apollo-cnt`:

```
docker build -t apollo-cnt .
```

To use Apollo we need to know where Apollo writes the config file in the container. With the cli in main (not the one on npm), you should be able to get it with: 

```
docker run apollo-cnt apollo config --get-config-file
/root/.config/apollo-cli/config.yaml # <- You should get this
```

To configure Apollo for, e.g., root access and to login use:

```
docker run --network host -v ~/.config/apollo-cli:/root/.config/apollo-cli apollo-cnt apollo config address http://localhost:3999
docker run --network host -v ~/.config/apollo-cli:/root/.config/apollo-cli apollo-cnt apollo config accessType root
docker run --network host -v ~/.config/apollo-cli:/root/.config/apollo-cli apollo-cnt apollo config rootCredentials.username admin
docker run --network host -v ~/.config/apollo-cli:/root/.config/apollo-cli apollo-cnt apollo config rootCredentials.password pass
docker run --network host -v ~/.config/apollo-cli:/root/.config/apollo-cli apollo-cnt apollo login
```

[...]

Issue #387, opened 2024-05-14
Create Dockerfile for image that runs CLI


garrettjstevens:
The dockerfile should be based on a node Docker image. It will run `yarn global install @apollo-annotation/cli`, and the CMD of the Dockerfile will run the CLI. Args passed to Docker when running the container based on this image will be passed to the CLI.

Issue #368, opened 2024-03-13
Publish packages to NPM


garrettjstevens:
We are going to try publishing under the `@apollo-annotation` scope. Things to keep in mind:

- We'll have to change the "name" of our packages that we publish, e.g. `jbrowse-plugin-apollo` will become `@apollo-annotation/jbrowse-plugin-apollo`
- Start with publishing `@apollo-annotation/jbrowse-plugin-apollo` and any of its dependencies.
- We'll have to remove `"private": true"` from the package.json of the packages we publish and replace it with
```json
{
"publishConfig": {
  "access": "public"
}
}
```
- We should probably match all of our published package versions, at least to start. I think `0.1.0` would be a good version for now.
- Make sure to use a tag when publishing (e.g. "beta") for now so we don't use the default "latest" tag until we're ready for a more official release. https://docs.npmjs.com/adding-dist-tags-to-packages
- It would probably be good to add some sort of single command to publish all the packages, maybe using `yarn workspaces foreach`
- There might also be some good tips here: https://yarnpkg.com/features/release-workflow

PR #393, opened 2024-05-22
Publish packages to NPM and publish the apollo-cli Docker image

dariober:
This PR should address #368, #387, and #391. It has two loosely related edits:

* Edit code to be publishable to npm (mostly renaming packages) and add workflow to publish packages (e.g. see [here](https://www.npmjs.com/package/@apollo-annotation/apollo-cli?activeTab=versions))

* Add Dockerfile to containerize the apollo-cli, add tests for docker, and add workflow to publish to GitHub Packages (see [here](https://github.com/gmod/Apollo3/pkgs/container/apollo-cli))


.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '18.x'
+          cache: yarn
+          registry-url: 'https://registry.npmjs.org'
+          scope: '@apollo-annotation'
+      - run: yarn --immutable
+      - run: |
+          yarn config set npmAuthToken ${{ secrets.NPM_TOKEN }}
+          yarn workspaces foreach --no-private --all npm publish --tolerate-republish --tag devel
```
garrettjstevens, line 23: There should be no tag specified here so that it gets the default release tag

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
...
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+  publish-docker:
+    name: Build and push Docker image
+    uses: ./.github/workflows/build-and-push-docker.yml
+    with:
+      tag: devel
```
garrettjstevens, line 30: No tag here, either

.github/workflows/publish.yml
```diff
@@ -0,0 +1,30 @@
+name: Publish Apollo
+
+on:
+  push:
+    tags: 'v[0-9]+.[0-9]+.[0-9]+*'
+
+jobs:
```
garrettjstevens, line 7: Need to add a job to deploy to demo server. The "Deploy to demo staging server" job from `push.yml` should be split into a separate shared workflow file and then called from both `push.yml` and here.

.github/workflows/push.yml
garrettjstevens, line 1: I think the "build" job's "if" statement needs to be updated to match the "deploy" job, since the "build" job uploads artifacts that the "deploy" job needs.

scripts/makeGitTag.ts
```diff
@@ -0,0 +1,119 @@
...
+  returncode: number
+  cmd: string
+  constructor(cmd: string, args: string[], verbose = true) {
+    this.cmd = `${cmd} ${args.join(' ')}`
+    if (verbose) {
+      process.stderr.write(`${this.cmd}\n`)
+    }
+    const child = spawn.sync(cmd, args)
+    this.stdout =
+      child.stdout === null ? '' : String.fromCodePoint(...child.stdout)
```
garrettjstevens, line 19: Use `stdout.toString()` instead of `String.fromCodePoint(...stdout)`

scripts/makeGitTag.ts
```diff
@@ -0,0 +1,119 @@
...
+  return packages
+}
+
+function checkTag(tag: string, testRegex = /v\d+.\d+.\d+.*/) {
+  if (!testRegex.test(tag)) {
+    throw new Error(`Invalid tag: '${tag}' does not match regex ${testRegex}\n`)
+  }
+}
+
+function updatePackageVersion(newVersion: string, packagesToUpdate: string[]) {
```
garrettjstevens, line 53: This should update all packages. Even if the private packages aren't getting published, their versions should still be updated. Also, it should use `yarn workspaces foreach` instead of calling it in each directory separately.

scripts/makeGitTag.ts
```diff
@@ -0,0 +1,119 @@
...
+  for (const line of lines) {
+    const j = JSON.parse(line)
+    const pack: string = path.join(j.location, 'package.json')
+    if (!fs.existsSync(pack)) {
+      throw new Error(`Error: Expected file '${pack}' does not exist\n`)
+    }
+    packages.push(pack)
+  }
+  return packages
+}
```
garrettjstevens, line 45: I don't think this function will be needed since `updatePackageVersion` should update everything and `git add` can use a glob instead.

scripts/makeGitTag.ts
```diff
@@ -0,0 +1,119 @@
...
+
+checkTag(tag)
+const packages = getPackages()
+updatePackageVersion(tag, packages)
+if (argv['update-only']) {
+  // eslint-disable-next-line unicorn/no-process-exit
+  process.exit(0)
+}
+new Shell('git', ['add', '--force', ...packages])
```
garrettjstevens, line 116: Can use a glob here, like `git add "*package.json"`

scripts/makeGitTag.ts
```diff
@@ -0,0 +1,119 @@
...
+new Shell('git', ['commit', '--message', m, '--', ...packages])
```
garrettjstevens, line 117: If files are already added, do they need to be re-listed in the commit command?

garrettjstevens (changes requested):

Issue #418, opened 2024-07-22
Add ability to define custom aliases for refSeq names


garrettjstevens:
We need to allow users to add aliases so that the ApolloRefNameAliasAdapter can return their custom aliases.

- Add `aliases` as an array of strings to `packages/apollo-schemas/src/refSeq.schema.ts`
- In `CollaborationServerDriver#getRefNameAliases`, add the values in `refSeq.aliases` to the `refNameAliases`

To allow users to add custom aliases:

- Create an `AddRefSeqAliasesChange` change with an interface that's something like this:
```js
new AddRefSeqAliasesChange({
  typeName: 'AddRefSeqAliasesChange',
  changedIds: idsOfChangedRefSeqs,
  aliases: [{id: refSeqId1, aliases: ["alias1", "alias2"]}, {id: refSeqId2, aliases: ["alias3", "alias4"]}],
})
```
- Create a menu item in `packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/addMenuItems.ts` called "Add reference sequence aliases"
- Create a dialog that opens when that menu item is clicked that allows people to add aliases to the refSeqs in an assembly
  - What I'm picturing for this is that the user can select an aliases file from their computer and then it will be parsed and a summary of the new aliases will be shown. Then when the user clicks "Submit", the `AddRefSeqAliasesChange` is submitted.
- Create a command that does something equivalent in the CLI

PR #419, opened 2024-07-30
Add ability to define custom aliases for refSeq names

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/418

image

Rendering of havana evidence track is working as expected after configuring the reference name alias.
image


garrettjstevens:
I think we can update the aliases on the client by doing something like this in `executeOnClient()`:

```ts
async executeOnClient(dataStore: ClientDataStore) {
  if (!dataStore) {
    throw new Error('No data store')
  }
  const { assemblyManager } = getSession(dataStore)
  const assembly = assemblyManager.get(this.assembly)
  assembly.configuration.refNameAliases.set(this.refSeqAliases)
}
```

PR #422, opened 2024-08-19
CDS check fix

shashankbrgowda:


packages/apollo-shared/src/Checks/CDSCheck.ts
```diff
@@ -1,277 +1,229 @@
...
+    feature?: AnnotationFeature,
   ): Promise<CheckResultSnapshot[]> {
-    await Promise.resolve()
-    return []
-    // if (feature.type === 'CDS') {
-    //   return checkCDS(feature, getSequence)
-    // }
+    if (!feature) {
+      return []
+    }
```
garrettjstevens, line 152: We need to operate on `featureSnapshot` for this check, since it has to run in the back end, too. You can probably copy some of the code, e.g. `cdsLocations`, over here to help, but you'll have to adapt it to work on a `featureSnapshot` instead.

garrettjstevens (changes requested):
Copied from email:

>  I still have issues with below features - CDS without any parent feature and mRNA without parent feature. For these features 'cdsLocations' returns empty locations.
> ```
> ---
> ctgA    bare_predicted  CDS     10000   11500   .       +       0       Name=Apple1;Note=CDS with no
> parent
> ---
> ctgA    predicted       mRNA    13000   17200   .       +       0       ID=cds-Apple2;Name=Apple2;Note=mRNA
> with CDSs but no UTRs
> ctgA    predicted       CDS     13500   13800   .       +       0       Parent=cds-Apple2
> ctgA    predicted       CDS     15000   15500   .       +       1       Parent=cds-Apple2
> ctgA    predicted       CDS     16500   17000   .       +       2       Parent=cds-Apple2
> ---
> ```

It's fine to just exit the check early with no errors in these cases, we'll decide more how to handle these types of features in the future.

PR #432, opened 2024-08-30
Add ability for indexed sequence file to be uploaded and used in an assembly

### LINKED ISSUE #405: Allow uploading file separately from adding assembly in CLI
Currently in the CLI, when adding an assembly from a file, the file is uploaded and then the file is used to add an assembly all in the same command. We should support a mode of operation where there is one command to upload a file, which returns a file ID, and another command to add an assembly using that file ID. This could help when people want to regularly clean out and re-add genomes so they don't have to re-upload the file each time.

### LINKED ISSUE #406: Add ability to upload a file and use it as a non-editable sequence for an assembly
Currently we have the options to
- upload a file (GFF3 or FASTA) and use it to load an assembly with sequence in the database
- use an external URL to create an assembly that doesn't store the sequence in the database

I think we need a third option:
- upload a file (FASTA) and use it to create an assembly that doesn't store the sequence in the database

The steps I see are:
- Allow user to upload either a FASTA/FAI or FASTA.GZ/FAI/GZI set of files.
- Store the file IDs in the assembly (modify assembly.schema.ts) similar to the `externalLocation` in the assembly (perhaps called `fileLocation`).
- In sequence.service.ts, check to see if the assembly has a `fileLocation` and if it does, fetch the sequence from the file similar to how the `externalLocation` fetches from a URL.

### LINKED ISSUE #407: Allow uploaded files to be already gzipped
Currently we can only upload uncompressed FASTA and GFF3 files. We should allow them to already be compressed.

To allow this:
- Add the `Content-Encoding: gzip` header when uploading a gzipped file (Content-Type will stay the same)
- Check the content encoding in the `FileStorageEngine` and if it's already gzipped, skip the compression step in the file upload

### LINKED ISSUE #424: Replace wrapLines with maxWidth
In the CLI code, there's a bunch of usage of `wrapLines` to wrap descriptions to 80 chars.

Instead of this, we should add `maxWidth` to our oclif help options:
```diff
diff --git a/packages/apollo-cli/package.json b/packages/apollo-cli/package.json
--- a/packages/apollo-cli/package.json
+++ b/packages/apollo-cli/package.json
@@ -70,6 +70,7 @@
   },
   "oclif": {
     "helpOptions": {
+      "maxWidth": 80,
       "flagSortOrder": "none"
     },
     "bin": "apollo",
```

This will have the same effect, but without us having to maintain consistent usage of `wrapLines` in all the command files.

garrettjstevens:
Fixes #405, fixes #406, fixes #407, fixes #424


packages/apollo-cli/src/commands/assembly/add-file.ts
garrettjstevens, line 1: This should be combined with the `add-fasta` command, they should not be separate commands.

Also, `add-fasta` and `add-gff` are unclear as to their function based on the command name. So what we need to do is:
- Rename `add-gff` to `add-from-gff`
- Rename `add-fasta` to `add-from-fasta`
- Combine `add-file` into `add-from-fasta`

Another thing to consider, maybe the argument to `add-from-fasta` should be positional and not a flag, since it's required in all cases. What do you think of this for example commands:
```sh
apollo assembly add-from-fasta ./my-genome.fa
apollo assembly add-from-fasta ./my-genome.fa.gz
apollo assembly add-from-fasta 507f1f77bcf86cd799439011
apollo assembly add-from-fasta ./my-genome.fa.gz --not-editable
apollo assembly add-from-fasta ./my-genome.fa.gz --not-editable --fai ./my-genome.fa.gz.fai --gzi ./my-genome.fa.gz.gzi
apollo assembly add-from-fasta 507f1f77bcf86cd799439011 --not-editable --fai 507f1f77bcf86cd799486197 --gzi 507f1f77bcf86cd785963292
```

packages/apollo-cli/src/commands/file/upload.ts
```diff
@@ -0,0 +1,76 @@
...
+      char: 'i',
+      description: 'Local file to upload',
+      required: true,
+    }),
+    type: Flags.string({
+      char: 't',
+      description:
+        'File type or "autodetect" for automatic detection.\n\
+        NB: There is no check for whether the file complies to this type',
+      options: ['text/x-fasta', 'text/x-gff3', 'autodetect'],
```
garrettjstevens, line 33: We need to add new options here, since we want to be able to tell the type later. Types I can think of are:
- text/x-fasta
- text/x-gff3
- application/x-bgzip-fasta
- text/x-fai
- application/x-gzi

packages/apollo-cli/src/fileCommand.ts
```diff
@@ -0,0 +1,107 @@
...
+
+    const init: RequestInit = {
+      method: 'POST',
+      body,
+      duplex: 'half',
+      headers: {
+        Authorization: `Bearer ${accessToken}`,
+        'Content-Type': type,
+        'Content-Length': String(size),
+        'Content-Encoding': contentEncoding,
```
garrettjstevens, line 79: Leave out `Content-Encoding` altogether unless it's gzip, maybe something like:
```ts
const headers = new Headers({
  Authorization: `Bearer ${accessToken}`,
  'Content-Type': type,
  'Content-Length': String(size),
})
if (isGzip) {
  headers.append('Content-Encoding', 'gzip')
}
```

packages/apollo-cli/src/fileCommand.ts
```diff
@@ -0,0 +1,107 @@
...
+    const body = pipeline(stream, progressTransform, (error) => {
+      if (error) {
+        progressBar.stop()
+        console.error('Error processing file.', error)
+        throw error
+      }
+    })
+
+    let contentEncoding = ''
+    if (file.endsWith('.gz')) {
```
garrettjstevens, line 67: Pass a param to `uploadFile` instead of checking the file name.

packages/apollo-collaboration-server/src/files/filesUtil.ts
```diff
@@ -61,9 +61,12 @@ export async function writeFileAndCalculateHash(
   })
 
   const fileWriteStream = createWriteStream(tmpFileName)
-  const gz = createGzip()
-  await pipeline(stream, gz, fileWriteStream)
-
+  if (originalname.endsWith('.gz')) {
```
garrettjstevens, line 64: Check the `Content-Encoding` instead of the filename.

packages/apollo-shared/src/Changes/AddAssemblyFromFileIdChange.ts
garrettjstevens, line 1: This should be folded in to the existing `AddAssemblyFromFileChange` instead of being a new change.

packages/apollo-cli/src/commands/assembly/add-fasta.ts
```diff
@@ -43,25 +36,32 @@ export default class Get extends BaseCommand<typeof Get> {
...
       char: 'x',
-      description:
-        'URL of the index. Required if input is an external source and ignored if input is a local file',
+      description: 'URL of the index. Required if input is an external source',
     }),
     force: Flags.boolean({
       char: 'f',
       description: 'Delete existing assembly, if it exists',
     }),
+    'no-db': Flags.boolean({
```
garrettjstevens, line 45: `no-db` should be something like `not-editable`, since that what is relevant to the user (the database is an internal detail they don't need to worry about)

packages/apollo-cli/src/commands/assembly/add-fasta.ts
```diff
@@ -43,25 +36,32 @@ export default class Get extends BaseCommand<typeof Get> {
...
+      char: 'n',
+      description: wrapLines(
+        "Do not load the fasta sequence into the Apollo database. \
+      This option assumes the fasta file is bgzip'd with `bgzip` and indexed with `samtools faidx`.\
+      Indexes should be named <my.fasta.gz>.gzi and <my.fasta.gz>.fai",
```
garrettjstevens, line 50: Auto-detection is good, but there should also be a way to explicitly pass the fai and gzi files.

garrettjstevens (commented):
One thing I didn't think about when we were discussing this before is that any `Change` can't have nodejs-specific code in it (i.e. no imports from `node:something`). This is because this code has to run  both in the browser and on the server. I think the nodejs-specific code that got introduced is what is causing the tests to fail.

What we'll need to do instead is that in `packages/apollo-collaboration-server/src/files/files.service.ts` we'll add a `getFileHandle` method that returns a file handle from a file document, very similar to how `getFileStream` in that same service file returns a file stream from a file document. Then the implementation of `LocalFileGzip` will be in collaboration server (e.g. in `packages/apollo-collaboration-server/src/files/filesUtil.ts`) instead of the shared code.

Then the code in the `Change` will look something like this:
```js
const { filesService } = backend
const { getFileHandle } = filesService
new IndexedFasta({
  fasta: getFileHandle(fastaDoc),
  fai: getFileHandle(faiDoc),
})
```
(Don't forget to add the definition for `getFileHandle` in `packages/apollo-common/src/Operation.ts`)

garrettjstevens:
@dariober the issue with the GZI filehandle we were seeing has been fixed

Issue #435, opened 2024-09-04
If you do "Add track to Apollo" multiple times in a session, it gets rid of the previously added tracks


garrettjstevens:
If you do "Add track to Apollo" multiple times in a session, it gets rid of the previously added tracks.

Found by @shashankbrgowda

PR #438, opened 2024-09-10
Save/Remove multiple tracks at once from track widget

shashankbrgowda:
There is a minor issue when we try to save multiple tracks into apollo in the same track widget window, newly added track replaces previous saved track so, only last track added will be saved to db. 

Fixes https://github.com/GMOD/Apollo3/issues/435


garrettjstevens:
Superseded by #445

PR #442, opened 2024-09-18
Fix converted GFF3 properties

garrettjstevens:
When combining multiple CDS into a single internal feature, we want to keep both IDs. This wasn't happening, so this PR fixes that.


garrettjstevens:
This PR also:
- Fixes a bug where the first registered user wasn't made an admin
- Re-enables the CDS checks

PR #444, opened 2024-09-19
Simplify the `getAccess` method in the CLI

garrettjstevens:
Since we can parse the flags in the base class, this does so to reduce the verbosity of `getAccess()`.

It also makes some fixes to get rid of some `eslint-disable` directives

### DIFF
--- a/.husky/pre-commit.js
+++ b/.husky/pre-commit.js
@@ -28,11 +28,6 @@ function main() {
     })
     spawn.sync(
       'yarn',
-      ['workspace', '@apollo-annotation/cli', 'oclif', 'readme'],
-      { stdio: 'inherit' },
-    )
-    spawn.sync(
-      'yarn',
       [
         'workspace',
         '@apollo-annotation/cli',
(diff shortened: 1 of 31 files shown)


garrettjstevens:
@dariober When I was removing some of the `eslint-disable` directives, I realized we were using `/dev/stdin` to read from stdin. Since that's not portable cross-platform (.e.g. on Windows), I changed it to use `process.stdin`, but had to make the `idReader` async to do so. I don't think it affected anything, but would appreciate you double-checking.

Also, should this be removed? From what I remembered, `.clirc` was what we used before we used the `conf` package.
https://github.com/GMOD/Apollo3/blob/28c2d3370356a0cf4caae644c80b1a9750d98fc5/packages/apollo-cli/src/utils.ts#L25

dariober:
@garrettjstevens Look good to me

> I changed it to use `process.stdin`, but had to make the `idReader` async to do so

I changed [edit.ts](https://github.com/GMOD/Apollo3/blob/6613297dbc1c5696e0f3081e4503767a1e5d48a0/packages/apollo-cli/src/commands/feature/edit.ts#L55) accordingly

> Also, should this be removed? From what I remembered, .clirc was what we used before we used the conf package.

Yes, it seems so. I removed it and simplified the checking if user is already logged in ([674e4ff](https://github.com/GMOD/Apollo3/pull/444/commits/674e4ffb836ba4323cce70131456ba316b2e071f)).

dariober (approved):

PR #455, opened 2024-09-26
Dockerfile for apollo collaboration server fails to build on Mac

shashankbrgowda:
The current Dockerfile for the collaboration server `ghcr.io/gmod/apollo-collaboration-server:devel` is not building successfully and does not run on a Mac machine. I updated the Dockerfile to use node:18-alpine as the base image and made some modifications to the build stages. These changes allow the image to build correctly and run on Mac environments without issues.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 # syntax=docker/dockerfile:1
 
-FROM node:18-alpine AS setup
+FROM node:20-alpine AS setup
 WORKDIR /app
 COPY package.json yarn.lock .yarnrc.yml ./
 COPY .yarn/ .yarn/
@@ -8,7 +8,7 @@ COPY packages/ packages/
 RUN find packages/ -type f \! \( -name "package.json" -o -name "yarn.lock" \) -delete && \
 find . -type d -empty -delete
 
-FROM node:18-alpine AS build
+FROM node:20-alpine AS build
 WORKDIR /app
 COPY --from=setup /app .
 RUN yarn install --immutable
@@ -16,7 +16,7 @@ COPY . .
 WORKDIR /app/packages/apollo-collaboration-server
 RUN yarn build
 
-FROM node:18-alpine
+FROM node:20-alpine
 LABEL org.opencontainers.image.source=https://github.com/GMOD/Apollo3
 LABEL org.opencontainers.image.description="Apollo collaboration server"
 WORKDIR /app
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,19 +1,24 @@
 # syntax=docker/dockerfile:1
 
-FROM node:18-alpine AS build
+FROM node:18-alpine AS setup
 WORKDIR /app
-
 COPY package.json yarn.lock .yarnrc.yml ./
 COPY .yarn/ .yarn/
 COPY packages/ packages/
 RUN find packages/ -type f \! \( -name "package.json" -o -name "yarn.lock" \) -delete
 RUN find . -type d -empty -delete
+
+FROM node:18-alpine AS build
+WORKDIR /app
+COPY --from=setup /app .
 RUN yarn install --immutable
 COPY . .
 WORKDIR /app/packages/apollo-collaboration-server
 RUN yarn build
 
 FROM node:18-alpine
+LABEL org.opencontainers.image.source=https://github.com/GMOD/Apollo3
+LABEL org.opencontainers.image.description="Apollo collaboration server"
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


garrettjstevens (changes requested):
There are quite a few changes here, are all of them necessary or is it possible to do a more minimal set of changes? At minimum, the `LABEL` needs to be replaced since that's used by the GitHub container registry.

In particular, I'd like to keep all three build stages since it offers better caching, which helps when having to build multiple times. Another thing is that we should avoid copying the whole `.yarn` directory over from the build step, since it includes all the installed dev dependencies, which will increase the size of our published image (`yarn workspaces focus` installs only the dependencies and not the dev dependencies)

shashankbrgowda:
> There are quite a few changes here, are all of them necessary or is it possible to do a more minimal set of changes? At minimum, the `LABEL` needs to be replaced since that's used by the GitHub container registry.
> 
> In particular, I'd like to keep all three build stages since it offers better caching, which helps when having to build multiple times. Another thing is that we should avoid copying the whole `.yarn` directory over from the build step, since it includes all the installed dev dependencies, which will increase the size of our published image (`yarn workspaces focus` installs only the dependencies and not the dev dependencies)

Issue seems to be with the `node:18` base image. Image build fine when we change the base image to `node:18-alpine` and also reduces size to 408 MB from 1.36 GB. 

We get below error with `node:18` base image, when we try to build the image and crashes when we start the container.
```
 => [build 3/7] COPY --from=setup /app .                                                                                                                                                                                             0.3s
 => [build 4/7] RUN yarn install --immutable                                                                                                                                                                                         6.3s
 => [build 5/7] COPY . .                                                                                                                                                                                                             2.7s 
 => [build 6/7] WORKDIR /app/packages/apollo-collaboration-server                                                                                                                                                                    0.0s 
 => ERROR [build 7/7] RUN yarn build                                                                                                                                                                                                 3.4s 
------                                                                                                                                                                                                                                    
 > [build 7/7] RUN yarn build:                                                                                                                                                                                                            
3.292 Usage Error: The project in /app/package.json doesn't seem to have been installed - running an install there might help                                                                                                             
3.292 
3.292 $ yarn run [--inspect] [--inspect-brk] [-T,--top-level] [-B,--binaries-only] [--require #0] <scriptName> ...
------
Dockerfile:17
--------------------
  15 |     COPY . .
  16 |     WORKDIR /app/packages/apollo-collaboration-server
  17 | >>> RUN yarn build
  18 |     
  19 |     FROM node:18
--------------------
ERROR: failed to solve: process "/bin/sh -c yarn build" did not complete successfully: exit code: 1
```

garrettjstevens:
I'm worried that the switch to `node:18-alpine` is hiding something else that might be going on. Could you try a couple other things for me? Could you try building with `node:18.20.4` and `node:20` to see if those work?

shashankbrgowda:
> 18.20.4



> I'm worried that the switch to `node:18-alpine` is hiding something else that might be going on. Could you try a couple other things for me? Could you try building with `node:18.20.4` and `node:20` to see if those work?

I'm getting same error with `node:18.20.4` and yarn install is failing for `node:20`

```
6.218 ➤ YN0002: │ @apollo-annotation/shared@workspace:packages/apollo-shared doesn't provide tss-react (p1a585), requested by @apollo-annotation/common and other dependencies.
6.218 ➤ YN0086: │ Some peer dependencies are incorrectly met by your project; run yarn explain peer-requirements <hash> for details, where <hash> is the six-letter p-prefixed code.
6.218 ➤ YN0086: │ Some peer dependencies are incorrectly met by dependencies; run yarn explain peer-requirements for details.
6.597 ➤ YN0000: └ Completed in 0s 592ms
6.597 ➤ YN0000: ┌ Fetch step
------
Dockerfile:14
--------------------
  12 |     WORKDIR /app
  13 |     COPY --from=setup /app .
  14 | >>> RUN yarn install --immutable
  15 |     COPY . .
  16 |     WORKDIR /app/packages/apollo-collaboration-server
--------------------
ERROR: failed to solve: process "/bin/sh -c yarn install --immutable" did not complete successfully: exit code: 139
```

garrettjstevens:
That's really strange that it fails at a totally different place under `node:20`. Does `node:20-alpine` work? Also, what is the version of Docker that you are running?

cmdcolin:
h.9a

shashankbrgowda:
> That's really strange that it fails at a totally different place under `node:20`. Does `node:20-alpine` work? Also, what is the version of Docker that you are running?

Yes, `node:20-alpine` works fine.

I'm running docker version 26.0.0, build 2ae903e.

Issue #340, opened 2023-11-29
Add configuration options for track's graphical and table views


garrettjstevens:
User should be able to:
- Don't show table (separate from collapsing the table, just not show it altogether)
- Only show table (similar to above)
- Show table above graphical instead of below
- Split graphical and table displays into separate tracks

garrettjstevens:
These options should be in the track menu. An example of how they might be added is:

```diff
--- a/packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel/base.ts
+++ b/packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/stateModel/base.ts
@@ -119,6 +119,40 @@ export function baseModelFactory(
           .apolloSelectedFeature
       },
     }))
+    .views((self) => {
+      const { trackMenuItems: superTrackMenuItems } = self
+      return {
+        trackMenuItems() {
+          return [
+            ...superTrackMenuItems(),
+            {
+              label: 'Show graphical display',
+              type: 'radio',
+              checked: self.graphical && !self.table,
+              onClick: () => {
+                self.showGraphicalOnly()
+              },
+            },
+            {
+              label: 'Show table display',
+              type: 'radio',
+              checked: self.table && !self.graphical,
+              onClick: () => {
+                self.showTableOnly()
+              },
+            },
+            {
+              label: 'Show both graphical and text display',
+              type: 'radio',
+              checked: self.table && self.graphical,
+              onClick: () => {
+                self.showGraphicalAndTable()
+              },
+            },
+          ]
+        },
+      }
+    })
     .actions((self) => ({
       setSelectedFeature(feature?: AnnotationFeature) {
         ;(
```

(I just made up the action/state names for this example, they might need to be different)

Issue #451, opened 2024-09-24
Make it so table view does not auto-open when a feature is selected


garrettjstevens:
If a feature is selected, the table opens even if collapsed:

![table_auto_open](https://github.com/user-attachments/assets/53770476-f473-446c-b822-c791064e174d)

The table should stay collapsed unless explicitly opened.

garrettjstevens:
Turns out it also pops open when you click and drag the track. May have to do with this:

https://github.com/GMOD/Apollo3/blob/89ccf49549f3bd640f6470e170da5048f3815988/packages/jbrowse-plugin-apollo/src/LinearApolloDisplay/components/LinearApolloDisplay.tsx#L173-L175

PR #461, opened 2024-10-08
Add configuration options for visibility of track's graphical and table views

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/340#issuecomment-2372509842, https://github.com/GMOD/Apollo3/issues/451

image
image
image


garrettjstevens:
Does this address these points of #340:
> - Show table above graphical instead of below
> - Split graphical and table displays into separate tracks

If so, what does the UI for these actions look like>

garrettjstevens:
Ok, can you make an additional PR for the other points of #340? They should also add menu items in the same way as this PR does.

Issue #450, opened 2024-09-24
Add filtering to LinearApolloDisplay


garrettjstevens:
JBrowse has track filtering, and one way of accessing it is through the track menu. The UI for adding filters looks like this:
![image](https://github.com/user-attachments/assets/6486dd5c-86cc-4eb8-9813-d9a0fc3827a3)

Apollo needs something similar, although because the Apollo track configuration is dynamically generated, we can perhaps do something different from jexl.

Types of filters users have requested:
- Filter by feature type, but ontology-aware (e.g. show all genes and `is_a` related terms, such as `protein_coding_gene` and `pseudogene`)
- Filter by specific feature type (e.g. only `gene`s and not `pseudogene`s)
- Filter by a custom feature attribute

garrettjstevens:
This will be accessible by a track menu item, the same idea as the menu items added in #461. The onClick for that menu item will queue a dialog (similar to the menu items in `packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/addMenuItems.ts`). For now the dialog can be a strict feature type filter, with a drop-down to choose a feature type like the drop-down in `packages/jbrowse-plugin-apollo/src/components/AddChildFeature.tsx`.

PR #469, opened 2024-10-21
Add filtering to LinearApolloDisplay

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/450#issuecomment-2417320391

image

image


garrettjstevens (changes requested):
I've noticed that if a feature type is in the filtered list, but then is removed, the features of that type don't disappear immediately, you have to refresh the page. This is because those features are already in the `seenFeatures` and don't get removed. I don't think we should remove them from the `seenFeatures`, though.

Instead of of checking the filter in the `afterAttach` of `layout.ts`, I think it should be checked in the `featureLayouts` and `featuresMinMax` views. That way the features are still always loaded, just not always displayed, and changes to the filter should show up immediately without a page refresh.

garrettjstevens (approved):

PR #482, opened 2024-11-25
Show ontology loading progress in Job List

dariober:


garrettjstevens:
@dariober I think I've fixed the issue where the ontology was always loaded in the test. The `meta` section of the pre-loaded indexeddb databases didn't match what was expected, so it always tried to re-load it. I added the missing `meta.storeOptions.prefixes` to the json files and fixed the import of those files so it transformed it from an object to a map. That made it match what Apollo was expecting, so now it doesn't try to re-load the ontologies every time.

dariober:
@shashankbrgowda This PR should notify the user that an ontology (sequence or gene ontology) is being loaded. It should show a progress bar in the "Running job" window (see screenshot below). If the ontology is already loaded then nothing should appear.

image

packages/jbrowse-plugin-apollo/src/OntologyManager/index.ts
```diff
@@ -55,6 +57,42 @@ export const OntologyRecordType = types
...
+      }
+      const equivalents = self.equivalentTypes.get(typeOf)
+      if (!equivalents) {
+        void self.loadEquivalentTypes(typeOf)
+        return false
+      }
+      if (equivalents.includes(queryType)) {
+        return true
+      }
+      return false
```
shashankbrgowda, line 94: return equivalents.includes(queryType)

shashankbrgowda:
When we delete the SO Indexed db in chrome dev tools and then refresh the page then loading appears two times and one of it doesn't complete. I encountered this only in this scenario so, may be we can ignore this ?
image

When we have a sub feature of type transcript, for some reason it renders the gene glyph like this
image

When we have only exon features without CDS then genericChildGlyph is used to render the gene. I assume this is done and merged already, just flagging it just in case. 
image

Other than that SO loading widget works fine and is really helpful.

packages/jbrowse-plugin-apollo/cypress/e2e/editFeature.cy.ts
```diff
@@ -174,7 +174,7 @@ describe('Different ways of editing features', () => {
     cy.contains('td', '=CDS1').should('not.exist')
   })
 
-  it('Suggest only valid SO terms from dropdown', () => {
+  it.only('Suggest only valid SO terms from dropdown', () => {
```
shashankbrgowda, line 177: skip other tests ?

packages/apollo-mst/src/AnnotationFeatureModel.ts
```diff
@@ -127,7 +127,14 @@ export const AnnotationFeatureModel = types
       return false
     },
     get transcriptParts(): TranscriptParts[] {
-      if (self.type !== 'mRNA') {
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any
+      const session = getSession(self) as any
```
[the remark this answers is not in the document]
dariober: > const session = getSession(self) as unknown as ApolloSessionModel

I thought about that but I couldn't import `ApolloSessionModel` from `jbrowse-plugin-apollo/src/session`

packages/jbrowse-plugin-apollo/cypress/e2e/editFeature.cy.ts
dariober: No, `.only` was a temporary edit that I should have reverted.

PR #488, opened 2024-12-03
Close LocalFileGzip after reading it

garrettjstevens:
I got an error like this loading data on the demo server:

```
    Error: importFeatures failed — 422 Unprocessable Entity 
    ({"message":"Error: EMFILE: too many open files, open 
    '/data/uploads/05dc2a73502099956dce87e417ed886c'","error":"Unprocessable 
    Entity","statusCode":422})
```

This fix still needs testing to see if it solves that error.

PR #489, opened 2024-12-04
Make sure LocalFile handles get closed

garrettjstevens:
As described in #488, I was getting this error when loading data on the demo server:

```
    Error: importFeatures failed — 422 Unprocessable Entity 
    ({"message":"Error: EMFILE: too many open files, open 
    '/data/uploads/05dc2a73502099956dce87e417ed886c'","error":"Unprocessable 
    Entity","statusCode":422})
```

It turns out #488 did not fix this, so after some more digging I found the solution in this PR. It turns out we have to be a bit careful about making sure we close filehandles when using `LocalFile` so that we don't have a lot of opened files.

### DIFF
--- a/packages/apollo-collaboration-server/src/files/files.service.ts
+++ b/packages/apollo-collaboration-server/src/files/files.service.ts
@@ -86,8 +86,6 @@ export class FilesService {
     return fileStream.pipe(gunzip)
   }
 
-  private fileHandleCache: Record<string, GenericFilehandle | undefined> = {}
-
   getFileHandle(file: FileDocument): GenericFilehandle {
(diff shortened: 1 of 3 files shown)


cmdcolin:
not sure if it's relevant but I was considering modifying the way LocalFile works in https://github.com/GMOD/generic-filehandle/pull/114 to not maintain an open filehandle. this way it would close and re-open for each read. while it could result in slightly less performance (benchmark could be made) it would reduce the need to meticulously manage the open/closing of the file (it is implicitly opened....explicitly closed as in this PR)

garrettjstevens:
That could also work. One thing I considered was using an LRU cache of filhandles instead of closing each one, but I didn't try it since, like you mentioned, I wasn't sure what the performance benefit would be for that added complexity.

PR #494, opened 2024-12-16
Add images to quick-start annotation guide

dariober:


garrettjstevens:
Thanks @dariober! FYI, I had to switch the images to use markdown image style since Docusaurus didn't understand the `img` tag.

PR #517, opened 2025-02-10
Introduce cypress-image-diff for visual testing

dariober:


shashankbrgowda (approved):
Create new annotation is working as expected after SO changes. I tried to run the new cypress tests in my local but its failing for me (may be I'm doing something wrong)

Cypress test Error: Image difference greater than threshold: 0

dariober:
@shashankbrgowda Thanks! 

> Cypress test Error: Image difference greater than threshold: 0

It maybe because of this: 

https://github.com/GMOD/Apollo3/blob/26c45e255c9527142eebc55b3cc743b62d0319fc/packages/jbrowse-plugin-apollo/cypress/e2e/visualGeneModel.cy.ts#L24-L29

Unless you have already done so, could you try to comment out line 26 and enable line 29? (It's unfortunate one has to do this, I hope to find a better solution)

If it fails you can have look at the content of `packages/jbrowse-plugin-apollo/cypress/cypress-image-diff-screenshots` to get an idea of where the difference is.

Issue #522, opened 2025-02-20
Gene glyph display issues


shashankbrgowda:
Assembly - [Homo_sapiens.GRCh38.dna.chromosome.21.fa.gz](https://ftp.ensembl.org/pub/release-113/fasta/homo_sapiens/dna/Homo_sapiens.GRCh38.dna.chromosome.21.fa.gz)
Annotation - [Homo_sapiens.GRCh38.113.chromosome.21.gff3.gz](https://ftp.ensembl.org/pub/release-113/gff3/homo_sapiens/Homo_sapiens.GRCh38.113.chromosome.21.gff3.gz)

- Exon and CDS not rendered
Location: 21:14,473,556..14,654,675

![Image](https://github.com/user-attachments/assets/4f88acab-db95-4d7f-a6d8-dfc049ac1c28)

- ncRNA_gene not rendering using gene glyph (Fix is needed for this as majority of the feature types is ncRNA_gene)

![Image](https://github.com/user-attachments/assets/64b38c21-4154-4337-b926-d4e5ee73adc2)

- When apollo track first loads gyphs are rendered using GenericChildGlyph for a split second before changing to Gene glyph
-  Sometimes gene is not rendered when I go to the location by using search and when we are scrolling
![Image](https://github.com/user-attachments/assets/74e37116-09b0-4752-ac1f-a0283fe9dbcb)

After refresh it displays fine
![Image](https://github.com/user-attachments/assets/f54ac030-c9f1-4618-a513-9d05b34b7a54)

garrettjstevens:
After merging #518, the area in the first screenshot looks like this, which I think looks correct:

![Image](https://github.com/user-attachments/assets/6fe0f62a-0b7f-4d4b-a605-a1e0e319ffc6)

garrettjstevens:
Similarly the second screenshot looks like this:

![Image](https://github.com/user-attachments/assets/6e235e1b-1097-4774-a0df-c058928b983e)

PR #524, opened 2025-02-22
Feature layout updates and loading indicator

shashankbrgowda:
- Change featureLayouts data structure to store feature id instead of the object and access feature from seenFeatures using feature id
- If feature min max range is large, rows boolean array in feature layouts was consuming too much memory leading to crash. Its replaced with a data structure which maintains feature coordinate range for each row
- Clean up seen features outside certain boundary to save memory
- Add loading indicator

Fixes https://github.com/GMOD/Apollo3/issues/522


garrettjstevens (changes requested):
Awesome work on the layout algorithm and cleaning up unused data. That will really help things run more smoothly. 

I really like the idea of indicating that data is loading, too, but I think it'll work better with a different loading indicator. Currently if you scroll even a tiny bit, the loading indicator blocks the top row, which might be what you're looking at. I think if we use something like a [circular indeterminate](https://mui.com/material-ui/react-progress/#circular-indeterminate) loading in the top right of the track, it still provides feedback but is not as likely to disrupt what the user is looking at.

garrettjstevens (approved):

PR #540, opened 2025-03-10
Fix #506: Warnings do not disappear after fixing

### LINKED ISSUE #506: Warnings do not disappear after fixing a feature
**Describe the bug**

Warnings do not disappear from glyphs after having fixed what causes them

**To Reproduce**

Add assembly from this gff+fasta file https://raw.githubusercontent.com/GMOD/Apollo3/refs/heads/python_to_typescript/packages/apollo-cli/test_data/checks.gff

Right-click on the (only) CDS -> "Edit feature details" -> Set start to 4 and end to 24

Two warnings should appear as expected.

Reset the CDS coordinates back to start 16, end 27. The warnings are still there (see screenshot).

**Expected behavior**

After re-setting coordinates to 16-27 warnings should go away.

**Screenshots**

image

**Version information:**

Commit version b714fbf522c01f285783486a8b8213fcb4d915a0

```
Browse 2
2.18.0

JBrowse is a [GMOD](http://gmod.org/) project

© 2019-2022 The Evolutionary Software Foundation

External plugins loaded

    ApolloPlugin 0.3.1
```

dariober:
The problem is a mismatch between `name` in CheckResult.ts and `name` in Check.ts. In CheckResult.ts name is (was) something like "InternalStopCodonCheck", but in Check.ts it is something like "CDSCheck". So when calling `clearChecksForFeature` in check.service.ts, the query returns nothing because 'CDSCheck' != 'InternalStopCodonCheck'.

This fix sets `name = "CDSCheck"` in class CDSCheck and adds a property `cause` to CheckResult to store the cause of the failing check, e.g cause = InternalStopCodonCheck. Also add property `causes` to class `Check` to enumerate all the possible causes for the fail.

Note test in test.ts to be fixed: FIXME: Checks stay after invalid operation

### DIFF
--- a/packages/apollo-cli/src/test/test.ts
+++ b/packages/apollo-cli/src/test/test.ts
@@ -323,7 +323,7 @@ void describe('Test CLI', () => {
     fs.unlinkSync('test_data/tmp.fa')
   })
 
-  void globalThis.itName('FIXME Checks are triggered and resolved', () => {
+  void globalThis.itName('Checks are triggered and resolved', () => {
     new Shell(`${apollo} assembly add-from-gff ${P} test_data/checks.gff -f`)
     let p = new Shell(`${apollo} feature get ${P} -a checks.gff`)
     const out = JSON.parse(p.stdout)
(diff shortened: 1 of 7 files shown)


garrettjstevens:
I think this sounds reasonable. I'll review once it's marked as ready.

Issue #527, opened 2025-02-27
Transcript details panel "Structure" section should support non-coding transcript types


garrettjstevens:
When displaying e.g. ncRNA of an ncRNA_gene, currently the "Structure" section of the transcript details panel is blank.

This should instead show a list of exons and their coordinated that can be edited. So for example, the below display would not show any UTR or CDS lines, but instead a line for each exon. I think we also still need the splice acceptor and donor columns.

![Image](https://github.com/user-attachments/assets/d5fa87e6-789f-4ee6-a8c7-7c72a35fd17f)

PR #544, opened 2025-03-18
Transcript details panel updates 2

shashankbrgowda:


dariober (approved):
It looks good. A couple of comments:

* Regarding issue #527, this is what I get for this pseudogene. Is it expected?

image

* It is not immediately clear how to use the trim function.

shashankbrgowda:
> It looks good. A couple of comments:
> 
> * Regarding issue [Transcript details panel "Structure" section should support non-coding transcript types #527](https://github.com/GMOD/Apollo3/issues/527), this is what I get for this pseudogene. Is it expected?
> 
> image
> 
> * It is not immediately clear how to use the trim function.

Thanks @dariober 

- Its doesn't fix pseudogene display issue. Pseudogene is not displaying with gene glyph because it doesn't have is_a relationship with gene. This issue is about not displaying the exon coordinates structure section to be edited in "Edit transcript details" widget. It used to display Structure section only when feature type is mRNA and If CDS was present
image

- Should I add a tooltip description for the trim icon ?

dariober:
> Should I add a tooltip description for the trim icon ?

Yes, I think that would be nice!

garrettjstevens:
Merging as-is for now, please add tooltip description in follow-up PR.

Issue #546, opened 2025-03-19
Create new transcript annotation from Jbrowse track


shashankbrgowda:
- If there is a transcript without a gene parent in Jbrowse track, allow it to be copied to a destination gene
- Add gene symbol/name along with coordinate information

PR #550, opened 2025-03-19
Add ability to copy transcript from Jbrowse track into apollo track destination gene

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/546

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/components/CreateApolloAnnotation.tsx
+++ b/packages/jbrowse-plugin-apollo/src/components/CreateApolloAnnotation.tsx
@@ -138,7 +138,7 @@ export function CreateApolloAnnotation({
 
   const [parentFeatureChecked, setParentFeatureChecked] = useState(true)
   const [checkedChildrens, setCheckedChildrens] = useState<string[]>(childIds)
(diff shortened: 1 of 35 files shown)


packages/jbrowse-plugin-apollo/src/components/CreateApolloAnnotation.tsx
```diff
@@ -226,7 +315,7 @@ export function CreateApolloAnnotation({
                   onChange={handleParentFeatureCheck}
                 />
               }
-              label={`${annotationFeature.type}:${annotationFeature.min}..${annotationFeature.max}`}
+              label={`${getFeatureNameOrId(annotationFeature, apolloSessionModel)} (${annotationFeature.min}..${annotationFeature.max})`}
```
dariober, line 318: Suggestion: Use `{annotationFeature.min + 1}` and on line 337 `{child.min + 1}` so the coordinate system is 1-based and consistent with the UI.

image

dariober (approved):
It appears that if there is no available destination feature, the CREATE button is active but clicking on it does nothing. Perhaps give the user some feedback?

image

image

PR #551, opened 2025-03-20
Fix #475 - Allow height to be reassigned

### LINKED ISSUE #475: Cannot resize graphical editor when displaying also the table editor
You display both graphical and table editor as below. If you want to reduce (or increase) the amount of empty space in the graphical editor you should point to mouse at the boundary between the two windows and drag up or down to resize. At the moment, this has no effect. The mouse pointer changes to double arrow but the windows are not resized.

image

dariober:


shashankbrgowda (approved):
lgtm

PR #560, opened 2025-04-02
Add LinearApolloSixFrameDisplay

### LINKED ISSUE #218: Move six-frame display to non-block-based rendering
Make changes with the same idea as LinearApolloDisplay where the renderer passes the canvas refs to the display model and the display model draws on the canvas.

haessar:
Fixes #218 

Main things to look out for are the drawHover logic, where individual CDS are determined from mouse position, and all other CDS belonging to the same parent mRNA are highlighted.

There is currently no dragging behaviour, and this is an open question. The equivalent view in Artemis doesn't allow dragging of the CDS features to alter them (it can be modified in a feature editor view instead). Should we therefore replicate this or adapt to allow dragging of the CDS?

Still need to fix tooltip on hover, and there is a TODO to fix the getFeatureLayoutPosition method.

Apollo3_issue_218


haessar:
[Fix six-frame display](https://trello.com/c/RfGOVqQ9/2-fix-six-frame-display)

haessar:
Now fixed the tooltip bug.

haessar:
Also fixed the getFeatureLayoutPosition method.

garrettjstevens (changes requested):
Looking good! Just a few things and I think it will be ready to merge.

It looks like there's one TS error to fix:

```
(!) Plugin typescript: @rollup/plugin-typescript TS2322: Type 'Glyph | Glyph' is not assignable to type 'Glyph & Glyph'.
  Type 'Glyph' is not assignable to type 'Glyph & Glyph'.
    Type 'Glyph' is missing the following properties from type 'Glyph': getRowCount, getFeatureFromLayout, getRowForFeature
src/TabularEditor/HybridGrid/Feature.tsx: (135:13)

135             glyph: displayState.getGlyph(getTopLevelFeature(feature)),
                ~~~~~
```

It looks like reverse strand features are getting drawn too high. I think I can just see the bottom of them in this screenshot (lower view, with LinearApolloDisplay shown above for reference):

image

Here's the GFF3 I used to get that screenshot in case it helps for testing: [3.subset.gff.gz](https://github.com/user-attachments/files/19817006/3.subset.gff.gz)

Also, it looks like the glyphs don't draw at the top of the track anymore when you zoom in past a certain level:

image

packages/apollo-mst/src/AnnotationFeatureModel.ts
```diff
@@ -240,6 +240,48 @@ export const AnnotationFeatureModel = types
...
+          // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+          featureTypeOntology.isTypeOf(child.type, 'pseudogenic_transcript')
+        ) {
+          const { children: grandChildren } = child
+          if (!grandChildren?.size) {
+            return false
+          }
+          return [...grandChildren.values()].some((grandchild) =>
+            // eslint-disable-next-line @typescript-eslint/no-unsafe-call, @typescript-eslint/no-unsafe-member-access
+            featureTypeOntology.isTypeOf(grandchild.type, 'exon'),
```
haessar, line 279: One slight concern I have here, which was introduced in [this](https://github.com/GMOD/Apollo3/commit/72def4bef098f59907390dee147e1761b91eee18) commit, is that a gene feature with CDS but no exon features will now not be regarded as a gene. Was that a deliberate choice, or oversight?

haessar:
I've made the fixes that we talked about yesterday. Also brought the looksLikeGene method into line with the latest in main (see my Comment above). I noticed the LAD GeneGlyph had functionalized a bunch of the logic (e.g. drawExon, drawLine). Recommend I do similar for LASFD before we merge?

PR #561, opened 2025-04-02
Add gene and transcript presets to "Add new feature"

dariober:
Shortcuts to create gene or transcript with subfeatures.


garrettjstevens:
Could you change this so the new gene/transcript feature gets created all at once? So instead of an `AddFeatureChange` for the gene, then another for the mRNA, etc., it does a single `AddFeatureChange`, using the "children" of the `addedFeature` for the subfeatures. This will make the changelog more accurately reflect the actions taken.

PR #563, opened 2025-04-02
Fix rendering of pseudogenes

garrettjstevens:


garrettjstevens:
Now renders `pseudogene`s similarly to `ncRNA_gene`s, with exons but no CDS.

Before:
image

After:
image

PR #565, opened 2025-04-08
Improve transcript widget edit location actions

shashankbrgowda:
Fixes transcript widget location editing

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
+++ b/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
@@ -115,8 +115,8 @@ export const TranscriptWidgetEditLocation = observer(
           continue
         }
         if (isMin && oldLocation === child.min) {
-          newLocation = newLocation - 1
-          if (newLocation < 0) {
+          const newLoc = newLocation - 1
+          if (newLoc < 0) {
             notify('Start location should be greater than 0', 'error')
             return
           }
@@ -125,7 +125,7 @@ export const TranscriptWidgetEditLocation = observer(
             changedIds: [child._id],
             featureId: feature._id,
             oldStart: child.min,
-            newStart: newLocation,
+            newStart: newLoc,
             assembly,
           })
           changeManager.submit(change).catch(() => {
@@ -150,7 +150,7 @@ export const TranscriptWidgetEditLocation = observer(
       }
     }
 
-    async function handleExonLocationChange(
+    function handleExonLocationChange(
       oldLocation: number,
       newLocation: number,
       feature: AnnotationFeature,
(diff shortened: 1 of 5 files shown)


packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
```diff
@@ -145,11 +150,13 @@ export const TranscriptWidgetEditLocation = observer(
       }
     }
 
-    function handleExonLocationChange(
+    async function handleExonLocationChange(
```
garrettjstevens, line 153: Instead of making async, use `void changeManager.submit()`

packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
```diff
@@ -159,31 +166,146 @@ export const TranscriptWidgetEditLocation = observer(
...
+              changedIds: [child._id],
+              featureId: feature._id,
+              oldStart: child.min,
+              newStart: newLocation,
+              assembly,
+            }),
+          )
+
+          // Submit all changes sequentially. gene -> transcript -> exon
+          for (const change of changes) {
```
garrettjstevens, line 236: These should all be in a `changes` attribute of a single `LocationStartChange` instead of three different `LocationStartChange`s. So it would look something like this:

```js
new LocationStartChange({
  typeName: 'LocationStartChange',
  changedIds: ['feature1Id', 'feature2Id'],
  assembly,
  changes: [
    { featureId: 'feature1Id', oldStart: 500, newStart: 490 },
    { featureId: 'feature2Id', oldStart: 500, newStart: 490 },
  ],
})
```

Same for the `LocationEndChange`s.

garrettjstevens (changes requested):

garrettjstevens:
The logic for whether or not something was in a gene glyph being in `mouseEvents` makes it so non-gene features can't be dragged anymore. I've proposed an alternate solution in a branch here that makes the parent/child expansion more general and has the gene glyph-specific logic outside of `mouseEvents`: [555_click_drag_improvement](https://github.com/GMOD/Apollo3/tree/555_click_drag_improvement).

garrettjstevens:
I've split this into two PRs, this one for the transcript widget and another for click and drag actions

garrettjstevens:
I was testing with a feature that looks like this:

image

And when I changed the last exon's max to something too small (e.g. I changed 19200 to 100), it got stuck in a loop of submitting changes for that position.

shashankbrgowda:
> I was testing with a feature that looks like this:
> 
> image
> 
> And when I changed the last exon's max to something too small (e.g. I changed 19200 to 100), it got stuck in a loop of submitting changes for that position.

Thanks @garrettjstevens 

This should be fixed. Haven't pushed my changes yet

Issue #559, opened 2025-04-02
Add non-colored appearance options for LinearApolloDisplay sequence


garrettjstevens:
- Under "Appearance" track menu item
  - Add options to show/hide start/stop codons
  - Add option to view protein translation lines in black and white
- Currently sequence in LinearApolloDisplay is hidden when bpPerPx > 3, explore if the display is still useful when raising that limit

haessar:
This is how it looks with the sequence frames rendering up to 30 bpPerPx:

![Image](https://github.com/user-attachments/assets/8300e059-7b98-4485-bcf8-d9a910a615d5)

(so not particularly helpful). The codons also become barely visible at this zoom level.

haessar:
I personally think it loses its utility after about bpPerPx == 5:

![Image](https://github.com/user-attachments/assets/3c616710-6556-4f2a-9e37-259bf6e7dcaa)

garrettjstevens:
> Do the "protein translation lines" refer to these
Yes
> So the option would be to make that grey scale (like currently exists in the Reference sequence track)?
I was imagining all the rows being white with the stop codons marked in black, like in Artemis. But really I think the key is contrast, so anything that gives high contrast between the row and the stop codons on that row will work.
> The codons also become barely distinguishable at this zoom level.
We might need to make sure the codons are always at least 1 pixel wide. For example, when calling `fillRect(x, y, w, h)`, make sure the `w` is always at least 1. I think this combined with the high-contrast coloring will allow it to be useful at higher zoom levels under certain conditions, like when you're looking for long areas with no stop codons.

haessar:
I believe Artemis also uses green for the start codons. I had intended to mimic the Artemis style in the LASFD after it is merged (so black/green stop/start codons overlaid directly on the six frame rows). But yeah, happy to do as you describe here for the LAD as well.

That's a good tip r.e. the min 1 pixel-wide codons. I will have a go implementing something for that tomorrow.

haessar:
I've added a "Use high contrast colors" appearance option, which sets codons to black/green, and has white frames with light grey outlines:
![Image](https://github.com/user-attachments/assets/cb99a6c9-01e9-4795-aff3-82c2e72fb0e7)
You can see the hover outline of the gene exon contrasting nicely (imo).

haessar:
![Image](https://github.com/user-attachments/assets/4ce067a2-b6ca-44ba-9440-470f5a6f2269)

PR #568, opened 2025-04-16
Additional appearance options for LinearApolloDisplay

haessar:


haessar:
I believe this is ready to review now, based on the reqs outlined in #559

garrettjstevens:
Looks good! I also realized what was making the reactions not work and went ahead an fixed that to get rid of the `console.warn`s.

PR #574, opened 2025-04-24
Add ability to merge and split exons

dariober:
Merge a selected exon with a neighbouring one on the same transcript. Attributes from the neighbouring exon are concatenated to the selected one except attributes `_id` and `gff_id`. Then delete the neighbouring exon.

TODO: Implement `getInverse()` method.

image

image


dariober:
This PR implements the merging of an exon into an adjacent one on the *same* transcript. It also implements the splitting of an exon into two exons of equal length, still sitting on the same parent transcript. The undoing these operation should also work. These options are available by right-clicking on the exon to be merged or split.

PR #576, opened 2025-04-30
Show details widget when we click on feature in gene glyph

shashankbrgowda:
- Show transcript details widget when we click on transcript/pseudogenic_transcript feature
- Show feature details widget when we click on cds/exon feature


dariober:
I tried it works great. I guess it was discussed and agreed that it should work only for transcript/pseudogenic_transcript and cds/exon?

shashankbrgowda:
> I tried it works great. I guess it was discussed and agreed that it should work only for transcript/pseudogenic_transcript and cds/exon?

This was requested by Havana team to open details widget when they click on features (cds, exon, transcript) in gene glyph. I think I had discussed with @garrettjstevens about this. 

Transcript and pseudogenic_transcript are both same, pseudogenic_transcript is explicitly added because it's not recognized as a transcript.

I have not implemented this for GenericChildGlyph and BoxGlyph.

PR #591, opened 2025-05-20
Alternate transcripts filtering for LinearApolloSixFrameDisplay

haessar:
Right click a gene feature (middle row), select "Filter alternate transcripts",  and then deselect any transcript IDs you want to de-render.


haessar:
Right-clicking on "gene" feature in the central rows brings up new context menu item "Filter alternate transcripts"
image

Also allowed toggling visibility of transcripts from table display, by right clicking on a "transcript" feature
image

PR #604, opened 2025-05-30
Add "Merge transcripts" action

dariober:
Current known issues:

1) When merging a feature (e.g. an exon) into the receiving feature, the children of that exon are lost. Basically, we extend the coordinates of the receiving feature to accommodate the incoming feature and we merge their attributes, but we don't account for children of the incoming feature.

2) Merging attributes: if an attribute of the incoming feature has a list of values with duplicates, the duplicates are removed and only distinct values are merged. Should we avoid de-duplicating altogether?

3) This merging of CDS is probably unexpected (part of the UTR should become an intron?):

Before:
image
After:
image


garrettjstevens:
I moved the menu item to the transcript sub-menu to match the recent menu changes we've done.

PR #608, opened 2025-06-04
Add open transcript details option on right click of child feature

shashankbrgowda:
When the transcript contain single exon thats the whole length of the transcript then we cannot open transcript details widget. This PR adds "Open transcript details" menu option when we right click on transcript or any of its child features.


garrettjstevens (approved):

garrettjstevens:
Adjusted after #618 so this is now in the transcript-specific context menu.

PR #616, opened 2025-06-25
split docker build action by platform arch

amorison:
Building the `aarch64` container directly on this architecture instead of cross-platform improves performances dramatically (about 3 minutes for the build, comparable to `x86_64`, instead of 50).

See an example run here (I tweaked the action to be able to run it on my fork for testing purposes): https://github.com/amorison/Apollo3/actions/runs/15873809734

Note: it would be good to have some kind of smoke test for the image before pushing it instead of building it and pushing it in one go. Is there an existing one we can easily set up as part of this PR, or should I open an issue for now?

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -20,14 +20,15 @@ on:
 jobs:
   build-and-push-docker:
     name: Build and push Docker image
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner:
+          - ubuntu-latest
+          - ubuntu-24.04-arm
     steps:
       - name: Check out
         uses: actions/checkout@v4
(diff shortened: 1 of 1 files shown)


garrettjstevens:
It's good to see the build being that much faster on native architecture. Now it just needs the merge step to make it into a multi-platform image.

Let's make an issue for the smoke test for now. My first thought was to start the container and ping the `/health` endpoint, but to even get to that point the container needs to have a MongoDB instance it can connect to. It's probably worth working that out separately.

amorison:
I've added the merge step. How should I go about testing this action? There might be a need to iterate a few times before getting it right, so it would be ideal if I can trigger it myself (e.g. pushing to temporary repository in the meantime).

garrettjstevens:
I tested the published test image on amd and arm and it looks like it's working well. Thanks!

PR #639, opened 2025-07-16
Add havana plugin code

shashankbrgowda:
It'll be easier to have Havana plugin related code in Apollo code base instead of maintaining it in separate branch.  It has become difficult to build the plugin package separately for havana deployment Instead of utilising the github workflow.

It won't display anything similar to custom plugin code unless we add the plugin in jbrowse config

```
"plugins": [
    {
      "name": "Apollo",
      "url": "apollo.js"
    },
    {
      "name": "HavanaApolloPlugin",
      "url": "https://unpkg.com/havana-apollo-plugin@0.0.13/dist/havana-apollo-plugin.umd.production.min.js"
    }
  ]
```

Without plugin added in jbrowse config
image

With Havana plugin (HavanaApolloPlugin)
image

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
+++ b/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
@@ -1,4 +1,10 @@
 import { type AnnotationFeature } from '@apollo-annotation/mst'
+import {
+  AddFeatureChange,
+  DeleteFeatureChange,
+  FeatureAttributeChange,
+  TypeChange,
+} from '@apollo-annotation/shared'
 import styled from '@emotion/styled'
 import {
   type AbstractSessionModel,
@@ -58,7 +64,7 @@ export const ApolloTranscriptDetailsWidget = observer(
     model: ApolloTranscriptDetailsWidgetState
   }) {
     const { classes } = useStyles()
-    const DEFAULT_PANELS = ['summary', 'location', 'attrs']
+    const DEFAULT_PANELS = ['summary', 'location']
     const [panelState, setPanelState] = useState<string[]>(DEFAULT_PANELS)
 
     const { model } = props
@@ -79,6 +85,7 @@ export const ApolloTranscriptDetailsWidget = observer(
     const apolloInternetAccount = internetAccounts.find(
       (ia) => ia.type === 'ApolloInternetAccount',
     ) as ApolloInternetAccountModel | undefined
+    const token = apolloInternetAccount?.retrieveToken()
     const role = apolloInternetAccount ? apolloInternetAccount.role : 'admin'
     const editable = ['admin', 'user'].includes(role ?? '')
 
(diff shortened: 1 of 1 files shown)


packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
```diff
@@ -112,6 +119,45 @@ export const ApolloTranscriptDetailsWidget = observer(
...
+      () => null,
+      {
+        key: 'status',
+        feature,
+        session,
+        assembly,
+        TypeChange,
+        DeleteFeatureChange,
+        AddFeatureChange,
+        FeatureAttributeChange,
```
garrettjstevens, line 133: Instead of passing these changes, the plugin using this extension point should add the `@apollo-annotation/shared` package (it's published to NPM so anyone can use it) and import the changes from there.

packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
```diff
@@ -112,6 +119,45 @@ export const ApolloTranscriptDetailsWidget = observer(
       props,
     ) as React.ElementType<CustomComponentProps>
 
+    const BiotypesComponent = pluginManager.evaluateExtensionPoint(
+      'Apollo-BiotypesComponent',
+      () => null,
+      {
+        key: 'status',
```
garrettjstevens, line 126: If `key` is always `"status"`, does it need to be a prop?

packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
```diff
@@ -112,6 +119,45 @@ export const ApolloTranscriptDetailsWidget = observer(
...
+        feature,
+        session,
+        assembly,
```
garrettjstevens, line 129: I think you can get the assembly from `feature.assemblyId` instead of passing it separately.

packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
```diff
@@ -112,6 +119,45 @@ export const ApolloTranscriptDetailsWidget = observer(
...
+        token,
+      },
+    ) as React.ElementType
+
+    const HavanaAttributesComponent = pluginManager.evaluateExtensionPoint(
+      'Apollo-HavanaAttributes',
```
garrettjstevens, line 150: Let's name the extension points by where they appear instead of having a custom Havana name, e.g. `"Apollo-TranscriptDetailsLocationCustomComponent"`

garrettjstevens (changes requested):
I'm ok with adding more extension points to the transcript details widget, but I think we need to make them more generic so other projects can use them if they want.

packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/ApolloTranscriptDetailsWidget.tsx
shashankbrgowda: Latest package doesn't contain the recent FeatureAttributeChangeDetails changes. Is it possible to publish that package separately ?

garrettjstevens: Yeah, I'll work on getting a release out.

garrettjstevens: You can now use the released v0.3.7

garrettjstevens:
@shashankbrgowda I've made an alternate suggested PR here: https://github.com/GMOD/Apollo3/pull/649

Basically it adds extension points labeled by where they are in the UI, so that it's more generic for others to use. The `'Apollo-TranscriptDetailsCustomSaveComponent'` in this PR would become `'Apollo-TranscriptDetailsCustomComponent-AfterSequence'`. Also, `'Apollo-TranscriptDetailsCustomTypesComponent'` and `'Apollo-TranscriptDetailsCustomAttributesComponent'` would be combined into `'Apollo-TranscriptDetailsCustomComponent-InsideLocation'`. You can combine your two components by wrapping them in a parent component that's something like this:

```jsx
function ParentComponent({ feature, session }) {
  return (
    <>
      <CustomTypesComponent feature={feature} session={session} />
      <CustomAttributesComponent feature={feature} session={session} />
    </>
  )
}
```

It also removes the `token` prop, but you can still get it inside the component like this:

```ts
const { internetAccounts } = getRoot<ApolloRootModel>(session)

const apolloInternetAccount = internetAccounts.find(
  (ia) => ia.type === 'ApolloInternetAccount',
) as ApolloInternetAccountModel | undefined
const token = apolloInternetAccount?.retrieveToken()
```

PR #640, opened 2025-07-18
Keep all attributes of merged features

### LINKED ISSUE #631: Options to handle attributes when merging
**Is your feature request related to a problem? Please describe.**

The "merging transcripts" functionality concatenates the attributes from the second transcript to the first transcript. This may be a problem if there is a "transcript_id" attribute that the user expects to remain stable (e.g. for Havana).

**Describe the solution you'd like**

Perhaps the merge functionalities for transcripts (and exons) should present the user with radio buttons where for each attributes you can choose: "keep only values from first feature", "keep only values from second feature", "keep both" - default to "keep both".

**Describe alternatives you've considered**

Currently the user can still edit the attribute(s) of interest to keep the value(s) of interest after merging, but it may be error prone and inconvenient.

dariober:
When merging transcripts or exons, make a string of all the attributes in the second feature and add it the first feature under attribute key `merged_with`. Fix #631


garrettjstevens:
@dariober, it turns out the errors we were seeing in this branch were actually symptoms of another problem, fixed in #656

PR #641, opened 2025-07-22
Make the "hovered feature" available to all displays on the session

garrettjstevens:
On the session, there is `apolloSelectedFeature`, so that the selected feature can be synchronized across displays. The feature currently being hovered over, though, is tracked separately by each display. This makes `apolloHoveredFeature` an entry on the session, so that it can by synchronized across displays as well. This is in preparation for splitting the sequence rendering into a separate display.

Before this, the `apolloHover` object contained the feature, the top-level feature, and the glyph. However `apolloHoveredFeature` is a single feature. We can derive top-level feature from `feature.topLevelFeature` and the glyph from `displayModel.getGlyph`, though, so we don't lose any information.

@haessar I wasn't able to apply this to the LASFD because it adds `cds` to the `apolloHover` object, and I wasn't clear how to re-derive that from just the feature. Could you take a look and see if you can make that work?


garrettjstevens:
I (think I) fixed the problem where the LASFD didn't select the right CDS on hover. When I tried to fix the tooltips, though, I realized that the global hover information really needed the the bp of the hover as well to do things like render info on parts of CDSs. I added the bp to the hover info and the tooltips appear to be working again.

Note, on the reverse strand in the LASFD the tooltips appear to be in the wrong place, but the same thing appears to happen on main, so that's not a regression in this branch.

@haessar could you try the branch out with these changes?

haessar:
@garrettjstevens that appears to have done the trick, good job!

I'd spotted the issue with the tooltip rendering height a few days ago: https://github.com/GMOD/Apollo3/issues/646. I'll probably create a separate PR to fix it.

Issue #634, opened 2025-07-15
Navigate to previous/next exon


garrettjstevens:
User request: a button to navigate to the next exon in a transcript glyph at the same zoom level. For example, if you are looking at this:

<img width="1427" height="267" alt="Image" src="https://github.com/user-attachments/assets/7bf329d9-ceba-47e9-a06d-9260b575b3fd" />

It's unclear how far you might have to scroll to find the next exon, and if you zoom out to find it, you lose the zoom level you were using before.

We'll add this as a context menu button under "exon" with a label that's something like "Scroll to next exon".

<img width="631" height="454" alt="Image" src="https://github.com/user-attachments/assets/8402d725-9c24-40d6-84ee-6a44b6bbd3b5" />

Note, I initially thought to do this in a general way so that users could e.g. "Scroll to next gene", but upon consideration, that would be considerably more work since the next gene may be outside the view and not loaded yet, and our backend drivers can only query by range and can't do general searches like "the next feature of this type in this region". With an exon, though, we know all the exons will be loaded since we load the whole top-level feature at once, even if some parts of the feature are out of view.

dariober:
This is how it looks at the moment:

<img width="647" height="589" alt="Image" src="https://github.com/user-attachments/assets/2b2c847d-a9bf-4446-9021-5991ed9cda11" />

The icon indicates the direction you would scroll to go up/downstream. I think the naming "previous/next" could be confusing when working with flipped coordinates.

PR #671, opened 2025-08-29
Navigate to up/downstream exon

dariober:
For example, this is a zoomed-out view of a transcript:

image

Zoom in and use the "Go to downstream exon":

image

Now we are on the exon on the right in the zoomed-out view. The view contains the whole exon plus padding equal to 10% of the exon length:

image

The arrow icons point in the direction you would scroll to. So they are reversed if you are in flipped coordinates mode.


dariober:
This PR also implements #672 since there is some shared code with #634 and #672 turned out to be quite simple (at least for now).

PR #690, opened 2025-09-24
Add feature background customization docs

dariober:
The code for setting background is simplified compared to main to avoid overly complex expression. Only background for toplevel features (e.g. genes) can be set.


garrettjstevens:
@dariober When we met earlier we were wondering how common it is to have a non-coding transcript as the child of a gene. It turns out it's pretty common in a C. elegans assembly I got from WormBase. There are quite a few features with type `nc_primary_transcript` that aren't coding. So before this change they get a red background:

image

But after they don't:

image

I personally think it's fine to keep going with this change and only set the background for the top-level feature, since you can tell the transcript is non-coding by the lack of CDS feature on that line, but thought I'd point out an example to give you a chance to take a look.

PR #704, opened 2025-10-08
Add a featureId to each annotation feature

garrettjstevens:
IDs are often an important aspect of features, and there have been several requests for being able to manage the IDs. Currently Apollo doesn't treat a GFF3 feature ID attribute any differently from other attributes.

This PR introduces a change so that `featureId` is a top-level attribute of all `AnnotationFeature`s. That way it can be guaranteed to be singular (attribute values like the default ID one are technically arrays) and can be more easily tracked and edited. On export, this `featureId` is placed in the GFF3 ID attribute.

Draft for now as I still need to figure out the best way to leverage this change when doing things that potentially create/destroy IDs, like merging or splitting exons or transcripts.


garrettjstevens:
Also still need to add ability to query by featureId

PR #705, opened 2025-10-08
Add configurable INDEXED_IDS to add to MongoDB indexedIds

### LINKED ISSUE #630: Query by ID (or other attributes?)
There are likely times that users will want to query the backend by feature ID. We have a text search index, but that's probably not sufficient.

One option is to find a way to efficiently query by any attribute. Another is to add another field to the feature schema where people can specify some external identifier they want to query by, and index by that.

garrettjstevens:
Fixes #630 

This adds a new server configuration `ADDITIONAL_IDS`, which is a comma-separated list. During feature import or creation, any feature that has attributes matching one of the ID names will have the value of that attribute added to the feature's `indexedIds` in MongoDB.

Builds off the `indexedIds` concept introduced in #704.


garrettjstevens:
No longer depends on #704.

PR #711, opened 2025-10-09
address race condition where assemblies exist on the server but aren'…

alpapan:
this is a defensive check to fix two issues that caused errors due to a sync and/or race condition between mongo and local storage

`Error: Could not find assembly "68d645473455e153641748db" to add feature "68e7ad6ad1890725fdbfdab0"`
(when adding a feature to an assembly that it can't find in its local state)

and

`Error: [mobx-state-tree] You are trying to read or write to an object that is no longer part of a state tree. (Object type: 'ApolloInternetAccount', Path upon death: '/internetAccounts/0', Subpath: '', Action: '/internetAccounts/0.postUserLocation()')`
(client-side state management)


garrettjstevens:
Closing as this was previously fixed in https://github.com/GMOD/Apollo3/pull/723

PR #731, opened 2025-11-07
Render start/stop codons on LASFD

haessar:
Notes:
* New Appearance toggle, with start:off, stop:on by default:
image

* I've used high contrast colours (start: green, stop: black) to match those used in Artemis:
image

* I've ensured the refSeq is loaded when much further zoomed out (like Artemis) to facilitate codon rendering. This is now tied to the zoom threshold value (set in Appearance):
image


haessar:
You'll probably recognise a lot of the logic from `jbrowse-plugin-apollo/src/LinearApolloReferenceSequenceDisplay/drawSequenceTrack.ts`

Issue #736, opened 2025-11-11
Query backend with multiple IDs


garrettjstevens:
If you want to query data on multiple features from the backend, it would be good to be able to specify all the IDs in one request instead of multiple `features/:featureid`-style requests. Here's a possible stub for this endpoint in `features.controller.ts`

```ts
/**
 * Get multiple features by their ids. If `topLevel` is true and multiple of
 * the specified ids share a common parent, that parent is included only once.
 * @param featureIds - Comma-separated list of feature ids
 * @param topLevel - Whether to return the top-level parent feature instead of
 * the specified feature
 * @returns
 */
@Get()
getFeatures(
  @Query('featureIds') featureIds: string,
  @Query('topLevel', new ParseBoolPipe({ optional: true }))
  topLevel: boolean | undefined,
) {
  return this.featuresService.findByIds(featureIds, topLevel)
}
```

This would be exposed in the CLI and also available for extension developers to use (HAVANA has expressed a need for this in their extension).

PR #741, opened 2025-11-27
Add ability to query backend with multiple IDs

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/736


garrettjstevens (changes requested):
Could you make this a `Get` request instead of a `Post`, like the example in #736?

garrettjstevens:
I changed the naming of the endpoints a bit. I want to keep the existing one as `getFeatures` even though `getFeaturesByRange` is a bit more specific because `getFeatures` is a sort of JBrowse idiom that I'd like to keep consistent with.

PR #742, opened 2025-12-05
Add option to add HTTP proxy for OAuth

shashankbrgowda:
## Details

EBI Infra doesn't allow access to Internet from onprem k8s cluster by default. In order to connect to the Internet from the pod we can set HTTP_PROXY env variable. Even though we can connect to Internet from the pod, Collaboration node server isn't picking that variable due to which authentication isn't working (connection to google idp is failing). This PR should fix that issue.

### DIFF
--- a/packages/apollo-collaboration-server/src/utils/strategies/google.strategy.ts
+++ b/packages/apollo-collaboration-server/src/utils/strategies/google.strategy.ts
@@ -74,8 +74,8 @@ export class GoogleStrategy extends PassportStrategy(Strategy) {
       // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
       const oauth2 = (this as any)._oauth2
       if (oauth2) {
-        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
-        oauth2.setAgent = agent
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-unsafe-call
+        oauth2.setAgent(agent)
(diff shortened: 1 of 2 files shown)


packages/apollo-collaboration-server/src/utils/strategies/google.strategy.ts
```diff
@@ -62,6 +63,22 @@ export class GoogleStrategy extends PassportStrategy(Strategy) {
...
+    if (proxy) {
+      const agent = new HttpsProxyAgent(proxy)
+
+      // Access the internal OAuth2 instance and set the agent
+      // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment, @typescript-eslint/no-explicit-any, @typescript-eslint/no-unsafe-member-access
+      const oauth2 = (this as any)._oauth2
+      if (oauth2) {
+        // https://github.com/ciaranj/node-oauth/blob/master/lib/oauth2.js#L20
+        // eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+        oauth2._agent = agent
```
garrettjstevens:
```suggestion
        oauth2.setAgent(agent)
```

garrettjstevens (changes requested):
This looks fine, but it needs to be documented and standardized. This includes:

- Adding the `HTTPS_PROXY` to `packages/apollo-collaboration-server/src/app.module.ts` (and maybe renaming it to `OAUTH_HTTPS_PROXY` to clarify that it's not a proxy for the whole Express server).
- Documenting the option in `packages/website/docs/02-installation/04-configuration-options.md`.
- Using `configService.get` to access the setting instead of checking `process.env`.
- Adding the same thing to `packages/apollo-collaboration-server/src/utils/strategies/microsoft.strategy.ts`

PR #751, opened 2026-01-20
JBbrowse 4 fixes

garrettjstevens:
A couple things didn't work with the latest JBrowse v4. These would be good candidates for adding to Cypress tests (although they both involve mouse interactions on the track canvas, which we haven't been able to get working in Cypress in the past).

One was the display's sub-menu being inaccessible, the other was the LinearBasicDisplay context menu extension throwing an error because the display's contextMenuFeature was unexpectedly undefined.


garrettjstevens:
Closing in favor of #753

PR #754, opened 2026-02-02
Make frame row order consistent in LinearApolloSixFrameDisplay

garrettjstevens:
I noticed that the row order for the CDS glyphs was not the same as the order for the stop codons in the LinearApolloSixFrameDisplay on the forward strand (reverse strand was correct).

This standardizes the CDS glyph frame row order so it's the same as the stop codon row order, as well as the same order as the ApolloSequenceDisplay.


garrettjstevens:
Before:
image

After:
image

Issue #757, opened 2026-02-16
Select Apollo track feature when it is the result of a search


garrettjstevens:
JBrowse main branch now supports running an action after a search result is selected and the view navigates to that result. See docs here: https://github.com/GMOD/jbrowse-components/blob/99ff0bae9547805a85971aad1de9bd9815d88ba0/website/docs/developer_guides/extension_points.md#lineargenomeview-searchresultselected

This will involve adding to an extension point in `packages/jbrowse-plugin-apollo/src/index.ts`, probably something like this:

```ts
import type BaseResult from '@jbrowse/core/TextSearch/BaseResults'

pluginManager.addToExtensionPoint(
  'LinearGenomeView-searchResultSelected',
  (_, { session, result, model, assemblyName }) => {
    const trackId = result.getTrackId()
    if (trackId.startsWith('apollo_track_')) {
      // get the feature from the result and select it on the session
    }
    return _
  },
)
```

PR #760, opened 2026-02-24
Select Apollo track feature that is the result of a search

shashankbrgowda:
Fixes https://github.com/GMOD/Apollo3/issues/757

Note:
It highlights only gene feature as search results from backend returns gene documents. Should we look for child feature containing user query ?

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
+++ b/packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts
@@ -11,6 +11,27 @@ import type { AbstractSessionModel, UriLocation } from '@jbrowse/core/util'
 
 import type { ApolloSessionModel } from '../session'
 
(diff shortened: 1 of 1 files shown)


garrettjstevens (commented):
I've found this doesn't work when the feature being searched isn't loaded yet. It gives an error like:

```
Failed to resolve reference '697cf682d3671008f0276906' to type 'AnnotationFeatureModel' (from node: /session/apolloSelectedFeature)
```

Maybe we can add a new method to the session called something like `apolloSetEventualSelectedFeature` that uses `mobx`'s [`when`](https://mobx.js.org/reactions.html#when) to wait for the feature to be in the data store and then calls `apolloSetSelectedFeature`.

As for only returning the parent feature, we should probably change that as well. There's no way to get the backend to send only the child feature, so we should probably modify `mapBaseResult` in `packages/jbrowse-plugin-apollo/src/ApolloTextSearchAdapter/ApolloTextSearchAdapter.ts` to get the subfeature from the parent before setting the `locString` and `matchedObject`.

PR #783, opened 2026-04-29
Clarify location box of transcript widget

garrettjstevens:
Moves the "Translation" accordion below the CDS/exon locations so it's clear that they are not part of that accordion box. Also add "Exons" label to separate the CDS and exon locations.

**Before:**

image

**After:**

image

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
+++ b/packages/jbrowse-plugin-apollo/src/FeatureDetailsWidget/TranscriptWidgetEditLocation.tsx
@@ -960,61 +960,11 @@ export const TranscriptWidgetEditLocation = observer(
       <div>
         {cdsPresent && (
           <div>
-            <Accordion>
-              <StyledAccordionSummary
-                expandIcon={<ExpandMoreIcon style={{ color: 'white' }} />}
-                aria-controls="panel1-content"
-                id="panel1-header"
-              >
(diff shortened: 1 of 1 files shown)


cmdcolin:
Would it be possible to present it as cds and utr?

garrettjstevens:
> Would it be possible to present it as cds and utr?

I'm not sure what you mean.

cmdcolin:
i now see what is happening by looking at the screenshot again, 

the way the data is shown is similar to bed12 thickstart/thickend, so the cds is the position of start codon and end of stop codon.  this might not be clear to common gff notions where transcripts have potentially many cds subfeatures, not just one cds range. The thing I referred to was an alternative approach which would be be to show 'CDS subfeatures' and 'UTR subfeatures'. this would then not show exons at all


Using the current way of doing things proposed by this pr though it could help to convey some of this detail via explainer text, or even a little 'gene cartoon' that helps show what the user is seeing

Another thing that could help convey meaning is having indicators like the start codon AUG at the start (for start codon) and UAA at the end (for stop codon) ...this would be similar sanity check to the splice site markers.

garrettjstevens:
Got it. The transcript location UI was made in collaboration with HAVANA users based on how they found to be easiest to edit the locations in the way they wanted. It could use some clarification, though, like you said, for users who aren't used to the way it's displayed here. I particularly like the idea of displaying the start and stop codons, I've created a feature request for it here: https://github.com/GMOD/Apollo3/issues/786

PR #790, opened 2026-05-12
Fix reauthenticate prompt when local token has expired

garrettjstevens:
This change ensures that when the user's local token expires, it shows a prompt to reauthenticate. Previously this didn't usually work unless you refreshed the page, but now no page refresh is required.

Fixes #773

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
+++ b/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
@@ -19,6 +19,7 @@ import { ConfigurationReference, getConf } from '@jbrowse/core/configuration'
 import { InternetAccount } from '@jbrowse/core/pluggableElementTypes'
 import {
   type AbstractSessionModel,
+  type UriLocation,
   isAbstractMenuManager,
   isElectron,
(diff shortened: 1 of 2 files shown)


packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
```diff
@@ -78,12 +79,75 @@ const stateModelFactory = (configSchema: ApolloInternetAccountConfigModel) => {
...
       },
     }))
+    .actions((self) => {
+      const superGetFetcher = self.getFetcher
+      return {
+        getFetcher(loc?: UriLocation) {
+          const fetcher = superGetFetcher(loc)
+          return async (input: RequestInfo, init?: RequestInit) => {
+            const response = await fetcher(input, init)
+            if (response.status === 403) {
```
shashankbrgowda, line 100: Do we need to consider 401 as well ?

garrettjstevens: For the specific case of the token expiring our server returns 403, so we shouldn't need to worry about 401 here.

PR #792, opened 2026-05-13
Add ability to color features

### LINKED ISSUE #750: Interactively color features
There are a couple use cases I've heard for interactively coloring features (and I'm sure there are more):

- Coloring a transcript of interest so you can click around to other transcripts, tracks, etc. and be able to easily go back to the one you were working on. This could be especially important for genes with a lot of transcripts.
- Coloring features different colors to indicate some sort of status, e.g. one color for "done", another color for "in progress", another color for "come back to later".

There's also the idea of programmatically coloring features by e.g. some attribute (which we have in a basic form with gene/pseudogene coloring), and we may want to implement interactive coloring as a special case/shortcut for that.

garrettjstevens:
Fixes #750. First draft at interactively coloring features. See linked issue for more discussion, but useful for e.g. keeping track of which transcript you're focusing on while clicking on other features.

https://github.com/user-attachments/assets/7623558c-7822-4434-bac4-45d5947653dc

To consider:
- Stores color as a feature attribute called `apollo_color`. This will show up in e.g. exported GFF3 files. Should there be an option to filter it out?
- Should the attribute name storing the color be configurable?
- Since it's a feature attribute, it will change color for all users in a collaborative installation.
- Do we want to be able to search for colored features? Or features of a specific color?


shashankbrgowda:
@garrettjstevens Havana team is happy with this feature. This is exactly what they're looking for. Thanks

PR #793, opened 2026-05-19
Add date based filtering for changes endpoint

shashankbrgowda:
### Details

- Add createdAfter query parameter to filter changes by createdAt
- Add  offset query parameter for pagination

PR #794, opened 2026-05-19
Add server-side sort, filter, and pagination to changelog

garrettjstevens:
This adds server-side sorting, filtering, and pagination to `GET` requests to the `changes` endpoint. This is then used to power the `DataGrid` from `@mui/x-data-grid`, making the UI much more responsive when there are many changes.

The changelog also still works for local editing when the changes are stored in indexedDb.

### DIFF
--- a/packages/apollo-cli/src/commands/change/get.ts
+++ b/packages/apollo-cli/src/commands/change/get.ts
@@ -26,14 +26,14 @@ In such cases you need to use the assembly ID.'
 
     const access = await this.getAccess()
 
-    const changes: Response = await queryApollo(
+    const response: Response = await queryApollo(
       access.address,
       access.accessToken,
       'changes',
     )
-    const json = (await changes.json()) as object[]
+    const { changes } = (await response.json()) as { changes: object[] }
 
-    let keep = json
+    let keep = changes
     if (flags.assembly !== undefined) {
       keep = []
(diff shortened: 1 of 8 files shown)


garrettjstevens:
@shashankbrgowda This is a proposed alternative to #793. I wanted to make sure any filtering and pagination we added would also work well with the changelog component in the UI. Datetime filtering and pagination is still available in this PR, just implemented a bit differently to work with the `DataGrid` component. Here's an example request:

```
http://localhost:3999/changes?assembly=697d2a056cf6618c23df2bb2&page=0&pageSize=15&sortField=sequence&sortOrder=desc&startTime=2026-05-12T06%3A00%3A00.000Z
```

Let me know if this works for your use case.

shashankbrgowda:
> @shashankbrgowda This is a proposed alternative to #793. I wanted to make sure any filtering and pagination we added would also work well with the changelog component in the UI. Datetime filtering and pagination is still available in this PR, just implemented a bit differently to work with the `DataGrid` component. Here's an example request:
> 
> ```
> http://localhost:3999/changes?assembly=697d2a056cf6618c23df2bb2&page=0&pageSize=15&sortField=sequence&sortOrder=desc&startTime=2026-05-12T06%3A00%3A00.000Z
> ```
> 
> Let me know if this works for your use case.

Thanks @garrettjstevens . This should work fine

PR #802, opened 2026-06-17
Fix LGV block API usage after upstream rename to displayedRegionIndex

cmdcolin:
This addresses some breaking changes from jbrowse-components. Upstream jbrowse-components webgl-poc refactor removed block.parentRegion and changed it to just store 'displayedRegionIndex'. This concept was then applied broadly across several APIs. Instead of storing parentRegion, users can manually lookup view.displayedRegions[displayedRegionIndex]. This was try to to simplify and optimize the block generation code: it was running getSnapshot frequently on the region objects, etc.

Sort of similarly, we renamed bpToPx's `regionNumber` attribute to `displayedRegionIndex`. 
We could potentially add backcompat for the bpToPx thing, but it may not be worth it to add backcompat for parentRegion.

If it is desirable to maintain compat with v4 and v5 of jbrowse though in apollo, we could add backcompat layer for 1 or both these issues and close this PR


garrettjstevens:
Closing as this commit is now part of #823

PR #803, opened 2026-06-17
Avoid accessing global document in webworker to fix https://github.com/GMOD/jbrowse-components/issues/5582

cmdcolin:
We deleted the 'worker polyfill' as part of a fairly aggressive effort to 'simplify' things. it can cause breakages in cases like this where the apollo plugin globally created document.createElement from the worker. it might be good to fix on apollo side


garrettjstevens:
Closing as this commit is now part of https://github.com/GMOD/Apollo3/pull/823

PR #821, opened 2026-07-23
Jbrowse 5 updates

garrettjstevens:
The current main branch of JBrowse, which will be released as JBrowse 2 v5, has a lot of changes, especially those that go with the webgpu updates.

This branch is for updating Apollo to accommodate those changes. Since it will not be backward compatible with JBrowse 2 v4, this will likely be released as Apollo 3 v2.

Some changes may still be backported to Apollo 3 v1.


cmdcolin:
From checking Apollo's source against jbrowse-components `main`, two `@jbrowse/core` names it imports are gone in v5:

- **`getParentRenderProps`** (`@jbrowse/core/util/tracks`) — used in `LinearApolloSixFrameDisplay/stateModel/base.ts` and `LinearApolloReferenceSequenceDisplay/stateModel/base.ts`. It fed the renderer registry, which is gone along with server-side rendering; nothing in the GPU pipeline consumes render props.
- **`isContainedWithin`** (`@jbrowse/core/util`) — `apollo-mst/src/ApolloRefSeq.ts:122`. Dropped with no replacement.

Both removals are deliberate, with the reason for each recorded in `KNOWN_REMOVALS` (`packages/core/src/ReExports/abiPreviousRelease.test.ts`).

*(Edited: an earlier version also listed `BaseTooltip` — that was a false positive, it isn't in your source.)*

---

cmdcolin:
a change needed to rpcmanager


jbrowse-plugin-apollo has two lines to change, on main and on jbrowse_5:

if (!('on' in handle.client && handle.client.on))  →  if (!('on' in handle && handle.on))
handle.client.on('apollo', …)                      →  handle.on('apollo', …)
handle.worker.postMessage(…)                       →  handle.postMessage(…)   // ×3

sorry for the breakage. if needed we can revert but hopefully this helps simplify usage. the client was technically a private field e.g. https://github.com/GMOD/jbrowse-components/blob/main/packages/core/src/rpc/WebWorkerRpcDriver.ts#L19 so likely not something to access directly anyways

cmdcolin:
Followed up on the RPC handle note above by checking this whole branch against current jbrowse-components `main` (`8e56c5c01a`, ~5200 commits past the `6c3280fc0b` pack pinned here). Result is in #831, based on this branch — merge it, cherry-pick from it, or just read it as a list.

Because the branch pins jbrowse by tarball, a clean control was possible: the same source, unchanged, against its own pin and against current main.

| run | errors |
| --- | --- |
| this branch vs its own pin | **2** — both pre-existing, unrelated to jbrowse |
| vs current main, deps not aligned | **~500** |
| vs current main, deps aligned | **24** |
| after #831 | **0** (plus jest 13/13, 11 snapshots unchanged) |

**The ~500 is the part I'd flag hardest**, because none of those errors names its cause. Core moved its `@jbrowse/mobx-state-tree` and `@mui/material` ranges, and both cross the plugin boundary as *types*. Two resolved copies gives you `[$type]` missing from every `types.model` argument, and `startCodon`/`bases` missing from `Palette` — the latter because a module augmentation cannot reach a second copy of the module it augments. Nothing in either message says "you have two copies of this package". That's on us to put in the v5 release notes; a core dependency bump on one of those packages is as much an ABI event as a renamed export, and no check we have sees it.

**Your `handle.client` note was the important one**, and it was worse than a rename: it typechecked. `@jbrowse/core/rpc/WebWorkerRpcDriver` had no `exports` entry, so there was no `WorkerHandle` to import — but also the module carrying that point's `ExtensionPointRegistry` declaration was never in Apollo's program, so `addToExtensionPoint` fell to its untyped overload and inferred the parameter from the hand-written `{ client, worker }` shape. Both fixed upstream (`8e56c5c01a`); with the type imported, putting the old shape back now fails at the call site with "check the callback against its args/result types". The generalizable bit: a registry declaration only constrains a plugin whose program already contains the module it lives in, and `exports` decides which of the 24 declaration sites that can be.

The rest, briefly: `LinearGenomeView-searchResultSelected` is a notification point (`listenToExtensionPoint`); `Core-preProcessTrackConfig` callbacks can't narrow their own parameter, and the type they need is now exported (`606e9762df`); React 19's peer requirement is real and reaches `packages/website` through your project-wide `yarn constraints`; `Menu` lost `slotProps.transition.onExit`; and the `SessionWithAddTracks` *type* is gone though the guard function survives. `@jbrowse/display-ui` and `@jbrowse/add-track-core` are new in core's dependency closure and unpublished, so they need tarballs alongside `render-core` until 5.0.0 ships.

Two latent bugs turned up on the way, both predating v5: `LinearApolloSixFrameDisplay` read `apolloDragging` from a second component parameter React never passes, so the check-result badges never went `pointerEvents: none` while dragging; and `featuresMinMax` couldn't prove its accumulators because `seenFeatures` resolves to `any`.

Two things #831 does **not** fix, flagged rather than papered over: the tarballs are from a local build and want re-packing from a clean checkout (or deleting once 5.0.0 is on npm), and `yarn lint` still fails — 83 errors, all `no-unsafe-*` off the `any`-typed display state models, down from 100 but not a v5 problem.

---

garrettjstevens:
Closing in favor of #823

PR #825, opened 2026-08-05
Add configurable timeout to CLI

garrettjstevens:
Adds a `--timeout` flag to the CLI commands. Defaults to `1h`, can be useful to set higher for commands expected to run for a long time.

Also refactors all fetching to happen in the base command so the usage of the timeout didn't have to be spread throughout the code.


garrettjstevens:
Also includes some improvements to the CLI tests so they run in 8 minutes instead of 20 on CI.

PR #826, opened 2026-08-12
Add the admin menus once, not once per initialize()

cmdcolin:
`addTopLevelAdminMenus` runs inside `initialize()`, which can run more than once — the autorun in `afterAttach` disposes only after the two awaits that follow it, and deliberately does not on failure — and each extra run puts another "Admin" submenu in the Apollo menu, because a menu contribution is appended to a log the root model replays on every open rather than added to a set.

Menu calls generally want to live in `configure()` for that reason, since it runs exactly once per plugin manager; these can't, as the role isn't known until login, so they get an `adminMenusAdded` guard in the same shape as the `roleNotificationSent` one a few lines above.

https://claude.ai/code/session_01JXCNVXEZMYnkeTS6cjKYBy

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
+++ b/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
@@ -478,7 +478,10 @@ const stateModelFactory = (configSchema: ApolloInternetAccountConfigModel) => {
       }
       return { postUserLocation: debouncePostUserLocation(postUserLocation) }
     })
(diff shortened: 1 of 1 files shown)


cmdcolin:
Superseded by #828 if that lands — it removes the need for this flag by separating the once-only work from the retryable work. Take either, not both.

cmdcolin:
Closing in favour of #828, which fixes the same defect for both effects at once — it separates the once-only startup work from the retryable work, so neither the menu call nor the socket listeners needs a flag. The branch stays if anyone wants to compare.

PR #827, opened 2026-08-12
Add the socket listeners once, not once per initialize()

cmdcolin:
Same root cause as #826 but a worse symptom: `addSocketListeners` registers each handler as a fresh arrow function, so socket.io cannot recognize a repeat and it does not no-op the way the `addEventListener` calls at the end of `initialize()` do — and `initialize()` can run more than once, since the autorun in `afterAttach` disposes only after the awaits either side of it and deliberately does not on failure.

A second registration means every `COMMON` message is submitted to the change manager twice, every `connect_error` raises two notifications, and every reconnect fetches the missing changes twice, so this guards on `socketListenersAdded` alongside the socket it belongs to; the token check throws before anything is registered, so a run that failed on it still retries cleanly.

https://claude.ai/code/session_01JXCNVXEZMYnkeTS6cjKYBy

### DIFF
--- a/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
+++ b/packages/jbrowse-plugin-apollo/src/ApolloInternetAccount/model.ts
@@ -374,10 +374,24 @@ const stateModelFactory = (configSchema: ApolloInternetAccountConfigModel) => {
     }))
     .volatile((self) => {
       const { origin, pathname: path } = new URL('socket.io/', self.baseURL)
-      return { socket: io(origin, { path }) }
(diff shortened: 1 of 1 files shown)


cmdcolin:
Superseded by #828 if that lands — it removes the need for this flag by separating the once-only work from the retryable work. Take either, not both.

cmdcolin:
Closing in favour of #828, which fixes the same defect for both effects at once — it separates the once-only startup work from the retryable work, so neither the socket listeners nor the menu call needs a flag. The branch stays if anyone wants to compare.

PR #829, opened 2026-08-14
Update how Apollo server-side plugins work

garrettjstevens:
Server-side plugins haven't been used much in Apollo yet (really just custom authentication). This updates the server-side plugin architecture to allow the plugins to be able to do a lot more things.

An example of an server-side plugin using this system to manage user assembly access is forthcoming.


garrettjstevens:
Example plugin using this new architecture here: https://github.com/GMOD/jbrowse-plugin-apollo-user-assembly-management

PR #831, opened 2026-08-22
Update jbrowse_5 to jbrowse-components main (8e56c5c01a)

cmdcolin:
Brings `jbrowse_5` up to jbrowse-components `main` as of `8e56c5c01a`, ~5200 commits past the `6c3280fc0b` pack this branch currently pins. Based on #821 rather than `main`, so it can merge into that branch or be cherry-picked out of it.

**Verified**: `yarn tsc --build tsconfig.json` clean across every workspace, plugin jest 13/13 with all 11 snapshots unchanged. Not run: Cypress (needs the collaboration server), and no runtime/browser check.

### Why the dependency bumps are not optional

Aligning them is what takes the error count from ~500 to 24, and the ~500 is the part worth knowing about, because none of those errors names its cause. `@jbrowse/core` moved its `@jbrowse/mobx-state-tree` and `@mui/material` ranges, and both are packages whose *types* cross the plugin boundary. With two copies resolved you get `[$type]` missing from every `types.model` argument, and `startCodon`/`bases` missing from `Palette` — the second because a module augmentation cannot reach a second copy of the module it augments. Neither message mentions duplication.

```
@jbrowse/mobx-state-tree  ^6.0.0   -> ^6.3.0
@mui/material             ^9.2.0   -> ^9.3.1
@mui/icons-material       ^9.2.0   -> ^9.3.1
@mui/x-data-grid          ^9.10.0  -> ^9.11.0
mobx                      ^7.0.0   -> ^7.0.3
mobx-react                ^10.0.0  -> ^10.0.2
@emotion/{react,styled}   ^11.10.6 -> ^11.14.0
react, react-dom          ^18.2.0  -> ^19.2.0
```

React 19 is core's peer requirement now (`react: >=19.0.0`), and it reaches `packages/website` too — `enforceConsistentDependenciesAcrossTheProject` is project-wide, so one workspace left on 18 fails `yarn constraints` for all of them. Docusaurus 3.10.1 declares `^18.0.0 || ^19.0.0`, so that move is free.

Also: `@jbrowse/display-ui` and `@jbrowse/add-track-core` joined core's dependency closure and are not on npm yet (nor is `render-core`), so both join the tarball resolutions or the install 404s.

### What broke, and why some of it was invisible

- **`Core-extendWorker`** — this is the one that matters most, because it was a *runtime* break that typechecked clean. The handle is a `WorkerHandle` with its own `on`/`postMessage`; this branch declared its own `{ client: { on }, worker: Worker }` and called `handle.client.on(...)`, which is a TypeError. It typechecks either way, because `@jbrowse/core/rpc/WebWorkerRpcDriver` had no `exports` entry — so there was no type to import, *and* the module carrying the point's `ExtensionPointRegistry` declaration was never in the program, so `addToExtensionPoint` fell to its untyped overload and inferred the parameter from whatever the callback claimed. Fixed upstream in `8e56c5c01a`; with the import in place, putting the old shape back now fails at the call site. Thanks for the heads-up in #821.
- **`LinearGenomeView-searchResultSelected`** is a notification point now — `listenToExtensionPoint`, which joins async handlers instead of letting the last one registered replace the rest. The type says so and names the replacement.
- **`Core-preProcessTrackConfig`** callbacks may not narrow their own parameter; the local `JBrowseTrackConfig` declared `trackId`/`type` required, which `TrackConfigSnapshot` does not promise. That type is exported as of `606e9762df`.
- **React 19**: a ref callback may return only a cleanup function (8 sites — the `await Promise.resolve()` is preserved as `void Promise.resolve().then(…)`, since mobx#3728 is still why it's there), and `useRef<T>(null)` is `RefObject<T | null>` (2 sites).
- **`Menu`'s `slotProps.transition.onExit`** is gone from `CascadingMenuProps`. All three callers used it to clear their context-menu items, which their own `onClose` already does.
- **`SessionWithAddTracks`** — the `isSessionWithAddTracks` *value* survives for prebuilt bundles; the type does not, and `session.ts` imported the type.

### Two latent bugs it turned up

- `LinearApolloSixFrameDisplay` read `apolloDragging` off a second component parameter. React passes function components one argument, so it was always `undefined` and the check-result badges never went `pointerEvents: none` during a drag. It now comes off the model, as `CheckResultWarnings` already does on the other display.
- `featuresMinMax` could not prove its accumulators non-`undefined`, because `self.seenFeatures` resolves to `any` there and assigning `any` to a `number | undefined` narrows nothing. Annotating the two values it reads is the missing fact; the four ifs collapse to two. The `any` itself is the state model's and is untouched.

### Caveats

- **The tarballs are built from a local jbrowse checkout**, as the ones they replace were. `core` is exactly `8e56c5c01a`; the other six are that tree plus another session's uncommitted work, whose only public-barrel changes were two names added and one removed, none of which appears anywhere in this repo. Re-pack from a clean checkout before treating them as final, and drop them entirely once 5.0.0 is on npm.
- **`yarn lint` still fails**, as it did before this: 83 errors, all `@typescript-eslint/no-unsafe-*` where the display state models resolve to `any` — down from 100 on the same source against the previous pack, with no rule firing in a file where it did not fire before. Not addressed here; it's a state-model typing problem, not a v5 one.

---


cmdcolin:
Follow-up from upstream: the reason the `Core-extendWorker` break was invisible is now closed at the source, so it won't recur here or in any other plugin.

`PluginManager.ts` pulls in each of the six core modules that declare an extension point away from itself, with side-effect type imports, and tsc keeps those in the emitted `PluginManager.d.ts` as bare `import './x.ts'` — which is what carries them to an installed consumer rather than only to the jbrowse repo. Every plugin's `install`/`configure` takes a PluginManager, so every plugin now gets every core-declared point whether or not it imports the module the declaration lives in.

Verified end to end against this branch: with the explicit `WorkerHandle` import **removed** and the old `{ client: { on } }` shape put back, the call now fails on its own —

```
index.ts(333,21): error TS2769: No overload matches this call.
  Overload 1 of 2, '(extensionPointName: "Core-extendWorker",
    callback: (extendee: WorkerHandle, …) => WorkerHandle | Promise<WorkerHandle>)'
```

which is what it did not do before. There's also a test on the jbrowse side asserting the set, with a negative control, since six imports that look unused are exactly what a later cleanup deletes.

Nothing to change here — this PR's explicit `import type { WorkerHandle }` is still correct and reads better than relying on inference, and it's what makes the annotation possible. No new tarballs needed either; the fix changes what plugins *see*, not anything this branch calls.

---

garrettjstevens:
@cmdcolin I've actually been tracking the latest JBrowse main in the [typescript_6](https://github.com/GMOD/Apollo3/tree/typescript_6) branch (PR https://github.com/GMOD/Apollo3/pull/823), and these breakages (except for the recent `Core-extendWorker` one) have already been fixed there.

